PR #9, opened 2017-06-28
Correct MariaDB Client Package Name

dfornika:
The mariadb client package on EPEL is named 'mariadb'


apetkau:
Thanks @dfornika you're right.

By the way, I saw you were trying to contact us on IRC.  You are also welcome to email us too if you aren't getting a response :)

PR #13, opened 2017-12-01
Apetkau/travis ci

apetkau:
Added integration tests with https://travis-ci.org/phac-nml/irida.  Will run all tests in parallel.  Excludes documentation build testing for now.


apetkau:
Hold off on merging this now, apparently it's reporting tests passed even though some failed.

apetkau:
Okay @tom114 this is good to go now.  I had to fix two issues:

1. I forgot to store the proper return value when running maven in `run-tests.sh`.  Now tests will properly indicate failure or pass.
2. For some reason there was some tables not being reset by `TableReset.xml` in our UI tests.  I'm not sure how this was even working on our integration test server.  I added them in and it now all works.

README.md
```diff
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/phac-nml/irida.svg?branch=development)](https://travis-ci.org/phac-nml/irida)
```
tom114: I'm noticing on your fork this is showing "Build: Unknown" even though the tests passed.  Is this the right format?  Or is it because you're linking to the `phac-nml` repo and it's not merged yet?

apetkau: Thanks.  Yep, it's the right format.  It shows `unknown` because it's linked up the the development branch in the right now in the `phac-nml` repo which has no tests run on it yet.

.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+addons:
+  mariadb: '10.0'
+
+jdk:
+  - openjdk8
+
+cache:
+  directories:
+    - $HOME/.m2
+    - src/main/webapp/node_modules/
```
tom114: Is this cache maintained between test runs?  Or just for different phases of the same test?  If it's maintained between runs I'm thinking this should be removed from the cache.  This directory often needs to be blown away when there's changes to library versions.

apetkau: It's maintained between test runs.  If you have access to the travis account (which you should) you can see it/delete it at https://travis-ci.org/phac-nml/irida/caches.

I didn't think of having to remove the `node_modules` directory.  Is this only when you wish to test older code though?  That tends to be when I encounter this issue.  That is, would it ever come up if we are only ever testing newer and newer code?

I can remove it pretty easily though.

tom114: I'd vote to remove the `node_modules` cache but keep the `m2` cache.  The node stuff doesn't take overly long to download fresh.

apetkau: Removed in f0735ef7a3564e10d4d064a43c1558ddc335adb0

PR #17, opened 2018-01-09
Apetkau/doc tests

apetkau:
Adds strict Javadoc documentation enforcement as part of the travis tests.  Specifically documentation will be required for all classes/methods in `/src/main/java` with the following caveats:

* Only required for `public` and `protected` methods.
* Not required for basic "getter" and "setter" methods.
* Not required for `@Override` methods as it assumes the parent is documented.  Overridden methods should still be commented if they do something different than the parent method.
* Not required (but highly encouraged) for constructors.

Related to https://irida.corefacility.ca/irida/irida/merge_requests/1218

@dfornika this may affect some of your code.  After this is merged in please merge into your branch and verify that the doc tests run successfully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ cache:
   directories:
     - $HOME/.m2
 
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install jekyll
+
 install:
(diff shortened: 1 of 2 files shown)


tom114:
Hah.  I was just about to do this.  Thanks for jumping on it.

dfornika:
@apetkau It looks like there were some fairly significant changes to [pipeline.launch.js](https://github.com/phac-nml/irida/blob/development/src/main/webapp/resources/js/pages/pipelines/pipeline.launch.js) in [this commit](https://github.com/phac-nml/irida/commit/4d06304a599c379c70a7a6b128ddf091a4392b11). It might take me a little while to resolve the merge conflict.

PR #43, opened 2018-06-18
Update MentaLiST-0.1.9

dfornika:
- Includes bugfix for downloading enterobase MLST schemes
- Adds new tool for building Neighbour Joining tree from MentaLiST MLST results.


apetkau:
Could you also update this workflow? - https://github.com/dfornika/irida/blob/133069263f24feb6e837b2608ece6626b130251b/doc/administrator/galaxy/pipelines/test/mentalist/mentalist.ga

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
```diff
@@ -616,6 +617,9 @@ private void tree(AnalysisSubmission submission, Model model) throws IOException
 
 		Analysis analysis = submission.getAnalysis();
 		AnalysisOutputFile file = analysis.getAnalysisOutputFile(treeFileKey);
+		if (file == null) {
```
dfornika: This may not be the cleanest way to check if a tree was generated. It may be better to insert a check before the `tree()` method is called.

apetkau: This solutions works great. Thanks 😄.  I think it could be simplified further by adding an additional case to the `catch` statement in the code that calls this method https://github.com/dfornika/irida/blob/de56393ddd86b1da2f235ac443300c7e22b843cb/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java#L231.  That is:

```java
catch (IOException | NullPointerException e) {
  logger.error("Couldn't get preview for analysis", e);
}
```

That way you can remove throwing an additional exception here and just handling the original `NullPointerException`.  And this code will now also properly work for the other preview types.

apetkau: Looking over this code again, I think it's probably best to leave it as it is.  Adding `NullPointerException` to the `catch` statement will catch all `NullPointerException`s, even those unrelated to a missing tree.  We may want those to trigger a user-visible error instead of just hiding the tree.

So, you can leave this code as is.

apetkau:
This works great.  Thanks @dfornika. I'm wondering if you could also remove the version `0.1.8` workflow, since it's unreleased and so doesn't need to be kept in IRIDA?

apetkau:
Thanks. This looks great to me.  I will merge it in when tests pass.

apetkau:
In case you're wondering why builds are sometimes failing, it looks like our builds are taking too long and hitting the TravisCI limits - https://github.com/phac-nml/irida/issues/48.  Everything works on our internal build server.

PR #46, opened 2018-06-26
Order analysis outputs by irida_workflow.xml order

### LINKED ISSUE #45: Is it possible to reorder pipeline output files
We chatted about this during last week's hackathon. Would it be possible to order the list of output files that appear under "Preview" tab after the analysis completes?  For example, when the IslandViewer analysis pipeline completes, it would be ideal to show the final two output files at the top of the list (considering they are the most relevant).  It would make sense to base the order on when they were created (i.e. reverse order based on time they were created). It also came up that the order could be specified based on the order the outputs appear in the irida_structure.xml file.

peterk87:
Fixes #45


apetkau:
Could you add an entry in the `CHANGELOG.md` about this?

peterk87:
@apetkau updated the changelog in 298f8c01219f8d362ed57752cb03b54525220919

peterk87:
Fixing an issue with the initial preview of a JSON file in 5e151f36f3eee9564d3ba95d2fac841ae1e53e14

PR #190, opened 2018-11-19
Fix issue with flaky testing on the Project Line List Page

joshsadam:


src/main/webapp/resources/js/pages/projects/linelist/components/TableControlPanel/TemplatesPanel/TemplateSelect/TemplateSelectOption.jsx
```diff
@@ -32,7 +32,9 @@ export function TemplateSelectOption(props) {
     <React.Fragment>
       <div className="templates-option">
         <span
-          className="template-option--name"
+          className={`template-option--name ${
+            index === current ? "t-select-option" : ""
```
tom114: I'm assuming this is saying it'll set the `t-select-option` class if the element is the currently selected element?  I know there's selenium matchers to find the selected element in a list, would that not work here?

joshsadam: You are correct in that selenium matchers can select the first element (you can use any css selector).  What I preferentially like to do to is when I need something specifically for testing purposes, I create a class prefixes with the `t-` so when I go back I know it is a target for a test.  This makes it very clear as to what is occurring and that it is not a target of styles.

joshsadam: So hear it also makes testing very easy when you are waiting for a specific item to be clickable by selenium.

tom114: Yeah I remember you talking about using `t-*` classes for testing.  I think that's certainly a better idea than relying on classes that are being used for CSS styles.  

I'm just considering whether it's better to use the `t-` classes over other selenium selectors?  In this case we're adding a JS conditional (extra code) that's introducing a class (extra code) for something that doesn't have anything to do with what the user wants to do.  I don't think the `t-` class should always be the go-to tool.  They're great for cases when it's awkward to get an element another way.  If Selenium provides an easy selector to get it otherwise I think things like this add bloat to the prod code that provides no benefit to the user.

I know in this case it's just a ternary operator that won't really introduce any performance hit, but I'm just wondering if we're starting to overuse `t-` here.

joshsadam: Ok, removed in 52afbdef4e5fd8c9490d66efbe75daf07c0ceeca

tom114:
Looks like the same test failing with these changes.

joshsadam:
@tom114 I think that fixed it.  I added a check to make sure that the save template modal was hidden after a template was saved (this was the previous step to the one that was failing).

tom114:
> @tom114 I think that fixed it. I added a check to make sure that the save template modal was hidden after a template was saved (this was the previous step to the one that was failing).

Thanks.  I'm running the test once more before merging.

tom114:
Looks good.  Merging.

PR #191, opened 2018-11-19
Ag-grid update

joshsadam:
This updated ag-grid to the most recent release.  Benefit in doing this is ag-grid now has a specific community branch that does not have the bloat from the enterprise edition that we do not have a licence for.


joshsadam:
Looks like the UI test is failing based on those same flaky tests.  I will send it back to you @tom114 when it is passing.

PR #204, opened 2018-11-22
Cart page empty state

joshsadam:
screen shot 2018-11-23 at 8 30 00 am

Since we are switching to a full page to view the cart, we need to display something if there are no samples available.


apetkau:
This looks great to me.

PR #205, opened 2018-11-23
Remove unused angular dependencies

joshsadam:


src/main/webapp/pages/projects/project_samples_metadata_template.html
```diff
@@ -118,104 +118,6 @@ <h4 class="modal-title" th:text="#{visualization.phylogenomics.field-modal.title
 
 
 <th:block layout:fragment="scripts">
-    <script type="text/ng-template" id="create-template.tmpl.html">
```
joshsadam: This is really old code that I found when I was grepping for dependencies.  It is from the original create metadata template 2 versions ago.

apetkau:
This works, thanks.

Issue #185, opened 2018-11-16
Change default shovill contig length threshold to 500 bp for SISTR pipeline


apetkau:
Imported from GitLab issue #682.  Originally posted on 2018/10/05 09:35AM

apetkau:
Originally I was going to change the threshold to 500 bp (from 1000 bp), but decided to change to 1 bp instead (that is, eliminate filtering of contigs as much as possible).

PR #206, opened 2018-11-29
WIP:Update default sistr

### LINKED ISSUE #195: Update SISTR metadata table to contain cgMLST and antigen serovar values

apetkau:
Updates the SISTR pipeline (0.3.0) in a few ways:

1. Updates `shovill` tool revision in Galaxy toolshed, since the one we were using is now out of date. This does not update the `shovill` version, which is still `0.9.0`, so I have not updated the SISTR workflow version.
2. Decreases the default thresholds for keeping contigs in Shovill for SISTR to the minimum values (contig lengths of 1 with a min coverage of 1). This is to keep as much information as possible in the assembly for SISTR to work with.
3. Updated the SISTR metadata post-processing to store the **cgMLST serovar** and **antigen serovar** in the table. Includes code to update the existing SISTR metadata records to include this information.

Fixes #195 and #185


apetkau:
I am going to split this pull request up into two separate ones.

PR #216, opened 2018-12-11
Hotfix/fix pipeline metadata

### LINKED ISSUE #213: Metadata derived from pipelines will not properly update all data on new runs of the same sample
If you run a pipeline (e.g., SISTR) which updates the metadata table, IRIDA will create proper entries in the metadata table and associated with with the pipeline submission (e.g., serovar `Heidelberg`, pipeline id `1`).

If you then run this pipeline again, metadata will be updated in the table, but not completely.  In particular, the associated pipeline id will still remain as `1` and any blank metadata entries in the second run will leave the old metadata entries still in the table.

For example, looking through the audit tables for metadata entry with an id of `9`:

```sql
select distinct mea.id,mea.REV,mea.REVTYPE,mea.value,pmea.submission_id from metadata_entry_AUD mea inner join pipeline_metadata_entry_AUD pmea on mea.id = pmea.id where mea.id = 9;
```

```
+----+-----+---------+------------+---------------+
| id | REV | REVTYPE | value      | submission_id |
+----+-----+---------+------------+---------------+
|  9 |  21 |       0 | Heidelberg |             1 |
|  9 |  40 |       1 | Dublin     |             1 |
|  9 |  60 |       1 | Typhi      |             1 |
+----+-----+---------+------------+---------------+
```

That is, while the `value` is updated depending on the data uploaded to the sample, the `submission_id` remains the same.

apetkau:
Fixes #213 

1. Updates the metadata **merge** functionality for a sample so it will properly merge metadata derived from pipelines.
2. Updates saving of metadata for SISTR results so it will save empty metadata entries (so old/new results won't get mixed up in the linelist).
3. Adds tests for the above functionality.

Note, this is a **hotfix** so will have to be merged into both `master` and `development`.


src/main/java/ca/corefacility/bioinformatics/irida/model/sample/Sample.java
```diff
@@ -338,27 +338,30 @@ public void setMetadata(Map<MetadataTemplateField, MetadataEntry> inputMetadata)
...
-					MetadataEntry metadataEntry = metadata.get(entry.getKey());
-					metadataEntry.setValue(entry.getValue().getValue());
+				// if the metadata entries are of the same type, I can directly merge
+				if (originalMetadataEntry.getClass().equals(newMetadataEntry.getClass())) {
+					originalMetadataEntry.merge(newMetadataEntry);
 				} else {
-					// otherwise add the new entry
-					metadata.put(entry.getKey(), entry.getValue());
+					// if they are different types, I need to replace the metadata entry instead of merging
+					metadata.put(entry.getKey(), newMetadataEntry);
```
tom114: Let me understand this section a bit.  This is testing for a case where someone makes a metadata entry named "bob", then later a pipeline comes along and writes an entry named "bob".  Since the pipeline entry should be a different type, it needs to create a new entry.  Is that correct?

src/main/java/ca/corefacility/bioinformatics/irida/model/sample/metadata/PipelineProvidedMetadataEntry.java
```diff
@@ -43,6 +48,21 @@ public PipelineProvidedMetadataEntry(String value, String type, AnalysisSubmissi
...
 		return submission;
 	}
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void merge(MetadataEntry metadataEntry) {
+		checkArgument(metadataEntry instanceof PipelineProvidedMetadataEntry,
+				"You cannot merge a non-PipelineProvidedMetadataEntry into a PipelineProvidedMetadataEntry: metadataEntry=" + metadataEntry);
```
tom114: Can you move this check up to `MetadataEntry` and make it general for any subclass?  It should likely explode if any different class of MetadataEntry is being merged here.  Just for a 2nd level check to what you're doing within Sample.

src/main/java/ca/corefacility/bioinformatics/irida/pipeline/results/impl/SISTRSampleUpdater.java
```diff
@@ -93,8 +93,9 @@ public void update(Collection<Sample> samples, AnalysisSubmission analysis) thro
 
 				//loop through each of the requested fields and save the entries
 				SISTR_FIELDS.entrySet().forEach(e -> {
-					if (result.containsKey(e.getKey()) && result.get(e.getKey()) != null) {
-						String value = result.get(e.getKey()).toString();
+					if (result.containsKey(e.getKey())) {
+						Object valueObject = result.get(e.getKey());
+						String value = (valueObject != null ? valueObject.toString() : "");
```
tom114: I'm a bit confused about what's going on here with the ternary operator.  Is it just stopping `null` from being written to a value?  What is a case that there would be an entry in the `result` map but the value would be `null`?

src/test/java/ca/corefacility/bioinformatics/irida/model/sample/unit/SampleTest.java
```diff
@@ -0,0 +1,127 @@
...
+import com.google.common.collect.Sets;
+
+import ca.corefacility.bioinformatics.irida.model.sample.MetadataTemplateField;
+import ca.corefacility.bioinformatics.irida.model.sample.Sample;
+import ca.corefacility.bioinformatics.irida.model.sample.metadata.MetadataEntry;
+import ca.corefacility.bioinformatics.irida.model.sample.metadata.PipelineProvidedMetadataEntry;
+import ca.corefacility.bioinformatics.irida.model.sequenceFile.SequenceFilePair;
+import ca.corefacility.bioinformatics.irida.model.workflow.submission.AnalysisSubmission;
+
+public class SampleTest {
```
tom114: There's already a `ca.corefacility.bioinformatics.irida.model.SampleTest`.  Merge these methods into that class please.

src/main/java/ca/corefacility/bioinformatics/irida/model/sample/metadata/PipelineProvidedMetadataEntry.java
apetkau: I could, but then I couldn't call `super.merge()` in `PipelineProvidedMetadataEntry` (as it would fail since it's not the same class).  I could just repeat the code in `MetadataEntry` here though if that's what you want?

src/main/java/ca/corefacility/bioinformatics/irida/model/sample/Sample.java
apetkau: Yes, correct. If we did not create a new `PipelineProvidedMetadataEntry` in this case then the associated `AnalysisSubmission` would be lost when the metadata is updated (since `MetadataEntry` does not store this information).

apetkau: One question I had was whether or not I should delete the old `MetadataEntry` in this case, since now it's unassociated with a sample. I don't think a `MetadataEntry` can be associated with multiple samples right now.  Deleting would likely require moving all this code to a service method though and/or playing around with delete cascade functionality.

src/main/java/ca/corefacility/bioinformatics/irida/pipeline/results/impl/SISTRSampleUpdater.java
apetkau: Currently, `null` would never be written to a sample, it would just skip over the entry.  However, this becomes problematic when updating the **cgMLST Type**, since this can be `null`. Skipping over `null` does not make sense here since it will then just leave the old **cgMLST Type** in the linelist, which may now mismatch the updated **Serovar**.

This line, specifically, is blanking out the field (cgMLST Type) in case of `null` instead of writing `null` (this is so it matches what you normally get as output from SISTR, which is an empty field instead of `null` displayed).

src/test/java/ca/corefacility/bioinformatics/irida/model/sample/unit/SampleTest.java
apetkau: Somehow I must have missed that. Merged in 9987e4806b0996665cca2ca04e2cae2fc0d322fb

src/main/java/ca/corefacility/bioinformatics/irida/model/sample/metadata/PipelineProvidedMetadataEntry.java
tom114: Maybe the logic would need to be changed a bit.  Instead of 
```java
metadataEntry instanceof PipelineProvidedMetadataEntry
``` 
maybe, more like
```java
metadataEntry.getClass().isAssignableFrom(this.getClass())
```
That way a `PipelineProvidedMetadataEntry` could merge into another `PipelineProvidedMetadataEntry`, but not `MetadataEntry` into `PipelineProvidedMetadataEntry`.  I think that's the right order, you might need to invert `metadataEntry` and `this` in the above code block.

apetkau: Thanks. I did not know about that method. Added in a1620b243b15b2f35c38ec1c19e89520bb652380

src/main/java/ca/corefacility/bioinformatics/irida/pipeline/results/impl/SISTRSampleUpdater.java
tom114: Got it.  Thanks.

src/main/java/ca/corefacility/bioinformatics/irida/model/sample/Sample.java
tom114: At the moment a `MetadataEntry` is only associated with one sample, but that's probably going to change as we get into the onto stuff.  I've been thinking about a good way to handle this, whether we should have a job that runs to garbage collect dead entries or something.  I'm not sure I have a good answer to that question yet.

apetkau: No problem. It's not going to hurt anything to have some unassociated `MetadataEntry`s in our database. And this particular situation is unlikely to come up right now anyways (unless someone uploads metadata with same field names to a project before running pipelines).

src/main/java/ca/corefacility/bioinformatics/irida/model/sample/metadata/PipelineProvidedMetadataEntry.java
apetkau: I've fixed this in 6acff1306bc7a334fe3fbabbea691b8b605f6bcc.

However, I had to change the code to not use `isAssignableFrom` as it won't actually work (`MetadataEntry` is assignable from `PipelineProvidedMetadataEntry`).

Instead, I just directly check if classes are equal in `MetadataEntry` and don't call the superclass method in `PipelineProvidedMetadataEntry`.

apetkau: After thinking about this a bit more, I can actually use the `super.merge()` method **and** remove the class check in `PipelineProvidedMetadataEntry`, just not with `isAssignableFrom`. I have to use `Class.equals()`. Fixed in 1d63f2e9977d35341472d0993e0f5936ecabb4b5

tom114: I was going to ask you about that this morning.  This should work.  Thanks.

PR #232, opened 2018-12-19
WIP:Adding update to SISTR pipeline thresholds

### LINKED ISSUE #185: Change default shovill contig length threshold to 500 bp for SISTR pipeline

apetkau:
Updates the SISTR pipeline (0.3.0) in a few ways:

   1. Updates shovill tool revision in Galaxy toolshed, since the one we were using is now out of date. This does not update the shovill version, which is still 0.9.0, so I have not updated the SISTR workflow version.
   2. Decreases the default thresholds for keeping contigs in Shovill for SISTR to the minimum values (contig lengths of 1 with a min coverage of 1). This is to keep as much information as possible in the assembly for SISTR to work with.

Fixes #185


apetkau:
I am actually going to break this up even further.

PR #246, opened 2019-01-09
Ability to lazy load javascript components

joshsadam:
## Description of changes
Previously the context path was causing issues with lazy loading components since they are generated during the compile time of the javascript and not on deployment of the war file.

**Why this is important:** Pages are starting to get heavy with javascript, and I want to start bringing the initial page load time down. My intention is to have the initial javascript bundle loaded  should only contain enough to get the page set up.  From there components can be loaded either right away, or when they are needed.

Most of the changes relate to any anchor tag that contained a `#` or relative url.  These needed to be updated to either be empty for the former, or the full link for the later.

## Related issue
N/A

## Checklist

* [X] ~~Tests added (or description of how to test) for any new features.~~ N/A
* [X] ~~User documentation updated for UI or technical changes.~~ N/A


joshsadam:
Ok, so after talking with @tom114 I have decided that this was a poor idea.  It would break all use of relative URLs in IRIDA.  Closing merge request.

PR #263, opened 2019-01-24
Hotfix/collection date

### LINKED ISSUE #256: Syncing a sample does not set the correct collectionDate
## Describe the bug

If a remote machine has a sample with a **Collection Date** filled in, and this gets synced to a local machine, the local date will be one day behind the remote date.

| Time | Remote | Local |
|------|--------|-------|
| 1    | Sample A (Collection Date = Jan 10, 2019) | None |
| 2    | Sample A (Jan 10, 2019) | Sample A (**Jan 9, 2019**) |

1. Create a sample and set the **Collection Date**.
2. Sync the project the sample was created in.
3. Check the date of the sample in the synced project. It will be one day behind the original date.

## Expected behaviour

I expected the dates to be identical.

apetkau:
### Description 

This fixes an issue where, upon synchronizing a remote project, the **collectionDate** could be up to a day off from the original.  This makes two changes to fix the issue:

1. Adds `@JsonSerialize` and `@JsonDeserialize` annotations around **Sample.collectionDate**.
2. Adds a custom Json serialize/deserialize class `ca.corefacility.bioinformatics.irida.web.controller.api.json.DateJson` which should serialize/deserialize to the correct format **YYYY-MM-DD** (which was the format that was used previously).

Note: This is a `hotfix` and so should be merged into `master` and `development`.

### Testing

1. Create a new Project containing a new Sample
2. Set the **Collection Date** for the sample.
3. Setup a new Client and Remote Connection
4. Synchronize the Project you created.
5. Check the **Collection Date** for the new synchronized project, either in the UI or in the database.

The expected results will be:

#### Development/Master

1. Local Sample

    image

2. Remote Sample

    image

3. `Sample` (local, remote) in Database

    ```sql
    select id,sampleName,remote_status,collectionDate from sample where id in (29,30);
    +----+------------+---------------+----------------+
    | id | sampleName | remote_status | collectionDate |
    +----+------------+---------------+----------------+
    | 29 | SampleDev  |          NULL | 2019-01-24     |
    | 30 | SampleDev  |            49 | 2019-01-23     |
    +----+------------+---------------+----------------+
    ```

#### This branch (with fix)

1. Local Sample

    image

2. Remote Sample

    image

3. `Sample` (local, remote) in Database

    ```sql
    select id,sampleName,remote_status,collectionDate from sample where id in (29,30);
    +----+------------+---------------+----------------+
    | id | sampleName | remote_status | collectionDate |
    +----+------------+---------------+----------------+
    | 29 | SampleDev  |          NULL | 2019-01-24     |
    | 30 | SampleDev  |            49 | 2019-01-24     |
    +----+------------+---------------+----------------+
    ```

## Related issue

Fixes #256 

Issue #283 is related, but not fixed by this merge.

## Checklist


apetkau:
It looks like this works by modifying the format of the JSON field that the **collectionDate** is serialized into (from **YYYY-MM-DD** to **Unix Timestamp in UTC time**).

## Before update (development)

```
{
  "resource" : {
    "sampleName" : "SampleDev",
    "collectionDate" : "2019-01-24",
```

## After update (this branch)

```
{
  "resource" : {
    "sampleName" : "SampleDev",
    "collectionDate" : 1548309600000,
```

In theory, our current docs actually specify that the **collectionDate** should be a Unix timestamp

> Must be a valid date (in IRIDA, that is the number of milliseconds since the epoch).

https://irida.corefacility.ca/documentation/developer/rest/#samples

But, this could affect scripts which may have been relying on **Collection Date** to be of format **YYYY-MM-DD** (though I can't think of any that use **Collection Date** from the Sample right now).

apetkau:
I've fixed this so it serializes/deserializes to `YYYY-MM-DD`.  That is:

```
{
  "resource" : {
    "sampleName" : "SampleDev",
    "collectionDate" : "2019-01-24",
```

This should match the previous REST API behaviour.

apetkau:
I tried to add a test to verify that doing a `PATCH` request to the API will update the Date, but due to https://github.com/phac-nml/irida/issues/283 it looks like this is broken for any Dates, so I left this test out.

src/main/java/ca/corefacility/bioinformatics/irida/web/controller/api/json/DateJson.java
```diff
@@ -0,0 +1,62 @@
+package ca.corefacility.bioinformatics.irida.web.controller.api.json;
+
+import java.io.IOException;
+import java.sql.Date;
```
tom114: Why the `java.sql.Date` instead of `java.util.Date`?  I know `java.sql.Date` is a subclass of `util`, but since we use `util` everywhere else why not here?  I remember you mentioning something about the resolution of `sql` vs `util` but that should be handled by the `@Temporal(TemporalType.DATE)` on `Sample` when saving to the database.

apetkau: The use of `sql` vs `util` is handled by `@Temporal` when saving to the database, but not when serializing/de-serializing to JSON. In that case, it's defaulting to serializing/de-serializing as a `java.util.Date`, which actually has different behaviours depending on the version of the package (the version of jackson we have now is serializing as **YYYY-MM-DD**, but the newer version is serializing as a long (milliseconds since epoch)). But, someplace in the serializion process there is an issue with handling time zones, leading to it trying to convert the time of `java.sql.Date` (which is in UTC) to Winnipeg time.

I believe in theory I could deserializing to `java.util.Date` in this class (since I'm implementing the process myself), but I haven't tested to make sure this works.

My reason for de-serializing to `java.sql.Date` is that, when debugging, it makes it explicit the resolution we are storing in this object. That is `java.sql.Date` only stores down to the day (internally, it stores as milliseconds since epoch in UTC time, but sets hour/minute/seconds to 12:00 AM). I also wasn't sure whether or not if I returned `java.util.Date` from the de-serializing process, some other code someplace would then try to convert this value to the proper time zone depending on the locale and so mess up the value.

apetkau:
Thanks @tom114. I am still looking into this, but I can confirm that it occurs for me too.  It also occurs in `master`. However, it doesn't seem to occur for me if I run IRIDA using Tomcat.

The two possibilities I am working with are:
1. Browser caching issue.
2. Incorrect construction of some URL (making it try to go to `/irida/...` instead of `/...`, which would explain why it works in Tomcat).

I'm having a hard time figuring out exactly what's going on though, since it seems like it should be a browser caching issue, but it also seems to persist across browsers and after clearing cache.

apetkau:
I can confirm that the redirect/connection issue I was having with `/irida/...` was from the `server.base.url` config parameter. Setting that properly and everything works.

For the issue you encountered, the one other thing I didn't mention is I cloned the **irida** repository again (to get rid of cached node modules) and it worked.

tom114:
Confirmed the "Waiting..." thing is an issue with a cached version of a JS library from Bower or Node or Yarn or something.  Blowing away all the cached JS libraries from the webapps directory fixed it.  I'm going to create an issue to more easily clean up all those libraries.

PR #290, opened 2019-02-21
Webpack Code splitting

joshsadam:
## Description of changes
* Code splitting: An example is the tour component on the line list page.
  * You can see the code splitting work if you go to the line list page with the developer tools open to the network panel.  After the page is loaded, click on the tour button (the question mark).  A new network request is made and the tour begins!

## Related issue
N/A

## Checklist

* [X] User documentation updated for UI or technical changes.


src/main/webapp/resources/js/pages/projects/linelist/components/Toolbar/Toolbar.jsx
```diff
@@ -85,10 +86,14 @@ export class Toolbar extends Component {
...
-                isOpen={this.state.tourOpen}
-                closeTour={this.closeTour}
-              />
+              <Suspense fallback={<span />}>
+                {this.state.tourOpen ? (
+                  <LineListTour
+                    isOpen={this.state.tourOpen}
+                    closeTour={this.closeTour}
+                  />
+                ) : null}
```
deepsidhu85: Should we not display something to the user such as "loading", etc if the component has not been loaded yet rather than returning null?

src/main/webapp/webpack.config.dev.js
```diff
@@ -8,7 +8,7 @@ exports.config = {
       target: "localhost:8080"
     },
     overlay: {
-      warnings: true,
+      warnings: false,
```
deepsidhu85: Any specific reason to suppress webpacker warnings?

src/main/webapp/resources/js/pages/projects/linelist/components/Toolbar/Toolbar.jsx
joshsadam: So normally I would say yes, but in this case no.  The reason is that everything that is needed is on the page.  The downloaded code opens a modal so nothing should be displayed until that is loaded.

src/main/webapp/webpack.config.dev.js
joshsadam: So all warning were showing up and it was slowing development.  They still get sent to the console so you will see them there.

Issue #192, opened 2018-11-19
Update user "cart" to allow a greater number samples


joshsadam:
Based on the way the current cart system functions, when users add a large number of samples to the cart the interface becomes sluggish.

**TODOs**

* [x]  UI:
  * [x] Remove client side code for slide out cart
  * [x] Cart Page
* [x] Server:
  * [x] Add samples to cart
  * [x] Remove samples from cart
  * [x] Remove project from cart
* [x]  Fix Tests:
  * [x] CartControllerTest
  * [x] PipelinesPhylogenomicsPageIT
  * [x] AssemblyPipelinePageIT
  * [x] CartControllerIT

tom114:
Fixes #68

PR #299, opened 2019-03-01
Cart/pipelines page

joshsadam:
## Description of changes
Re-integrate testing for launching pipelines.

## Related issue
#192 

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/ria/web/pipelines/PipelineController.java
```diff
@@ -136,50 +136,6 @@ public PipelineController(SequencingObjectService sequencingObjectService,
 		this.iridaPipelinePluginStyle = iridaPipelinePluginStyle;
```
apetkau: It looks like `iridaPipelinePluginStyle` is no longer used in this class. Could it be removed?

joshsadam: Removed in d210b2c8a786a046301efaa93f37a2bbc0e3229a

apetkau: Thanks @joshsadam.

I just noticed there's two methods which are unused in this class too `extractPipelineParameters` and `getCartSummaryMap`. Could you check to make sure these aren't required and remove them if so?

joshsadam: Good catch, removed in aa2c631d9e276eb6e2b17542e8ce42304eac1a82

apetkau: Looks great. Thanks @joshsadam

PR #300, opened 2019-03-04
Hotfix/configure executor threadpool

### LINKED ISSUE #301: Slowness of submitting jobs in Galaxy
## Describe the bug

I have noticed that IRIDA can get into a state where jobs are still being submitted to Galaxy, but only very slowly. This mainly occurs if there are a lot of single-sample pipelines submitted (e.g., SISTR or Assembly). The status in IRIDA will say **Submitting** or **Preparing** and it won't change for a long time, despite the many resources given to Galaxy.

Link IRIDA to Galaxy backed by a large compute cluster and submit large numbers of SISTR or Assembly jobs. Despite the large resources provided by the cluster, only a small number of jobs will be scheduled at once (e.g., upload jobs).

## Expected behaviour

I expected IRIDA to be able to schedule many jobs at once if the cluster resources are available.

## Additional context

None.

apetkau:
## Description of changes

Added support to configure the number of threads used for workflow/analysis tasks in IRIDA (using `irida.workflow.analysis.threads`).

Note, this is a **hotfix** and so should be merged into both **master** and **development**.

## Related issue

Fixes #301 

## Checklist


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Changes
 --------------
 * [UI/Developer]: Updated to lodash v4.17.11 to fix security issue. (19.01.1)
 * [Admin]: Added message to add `irida.db.profile` param for Tomcat in docs and upgrading guide.
+* [Admin]: Made analysis task pool size configurable with `irida.workflow.analysis.threads`. (19.01.2)
```
tom114: Can you add something to the UPGRADING.md file as well?

> *something something* new config value *something* default value is 4 *something something* helps with lots of running jobs here's the docs.

src/main/java/ca/corefacility/bioinformatics/irida/config/services/IridaApiServicesConfig.java
```diff
@@ -139,6 +139,9 @@
 
 	@Value("${irida.debug.nrepl.server.port:#{null}}")
 	private Integer nreplPort;
+	
+	@Value("${irida.workflow.analysis.threads}")
+	private int analysisTaskThreads;
```
tom114: Something odd, if I set this to 0 it still runs the analysis tasks.  Any idea why?

apetkau: I am investigating this. I may want to rename the variable, since we aren't actually specifying a fixed number of threads here, just the `corePoolSize`, that is the number of threads that should be kept active at any given time even if no work is available.

>A ThreadPoolExecutor will automatically adjust the pool size (see getPoolSize()) according to the bounds set by corePoolSize (see getCorePoolSize()) and maximumPoolSize (see getMaximumPoolSize()). When a new task is submitted in method execute(Runnable), and fewer than corePoolSize threads are running, a new thread is created to handle the request, even if other worker threads are idle. If there are more than corePoolSize but less than maximumPoolSize threads running, a new thread will be created only if the queue is full. By setting corePoolSize and maximumPoolSize the same, you create a fixed-size thread pool. By setting maximumPoolSize to an essentially unbounded value such as Integer.MAX_VALUE, you allow the pool to accommodate an arbitrary number of concurrent tasks. - https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html

So, setting this to `0` just means that no threads will be kept active when no work is available, but more threads can be created even if there is more work.  But, what I'm confused about is why this wasn't just working before (e.g., why weren't we getting > 4 analysis tasks scheduled at once when the pool size is `4` if we can create more than this number of threads anyways).

CHANGELOG.md
apetkau: Done in ddcd8acf884453cc72dbeba9fc2e9557e4314855

src/main/java/ca/corefacility/bioinformatics/irida/config/services/IridaApiServicesConfig.java
apetkau: I've fixed this in bc691499597c9c30936b77e3bb8960c07fe1fdea and 5df737e842ee31f3ba2e22c39bc0c7012b33f56d.

I added a check if number of threads <= 0 and will throw an exception. I also switched the thread pool to a fixed thread pool (for some reason I was using a scheduled task thread pool before, but there is no need to schedule tasks here).

For why it was working with threads == 0, I think it goes like this. For a scheduled task thread pool, the threads you set corresponds to `corePoolSize`, with `maxPoolSize` set to unlimited. In theory, this should still allow some tasks to run if you overflow the queue (which will start scheduling more threads than `corePoolSize` up to `maxPoolSize`). But, the queue is also defaulted to unlimited so this will never happen (this is likely something that the scheduled task pool constructor should have checked for).

For our purposes though, we want a fixed thread pool (e.g., fixed number of threads available for handling tasks with rest of tasks queued up). So, I made that change.

tom114: I think that's a good change.  Thanks for checking it.

Issue #240, opened 2019-01-02
Add ability to run any analysis pipeline on data upload


tom114:
## Describe your idea for a new feature
Any enabled analysis pipeline should be able to run automatically when data is uploaded.  Currently we allow SISTR or Assembly through checkboxes in the project settings and parameters are not customizable.  Users should be able to set any pipeline and customize the parameters they want to use.

* [x] Rework pipeline gui to allow saving templates
* [x] Update GUI to show automated pipelines in project settings page
* [x] Run automated pipelines on data upload 
* [x] Replicate old assembly and sistr saving features in new file processor
* [x] DB upgrade to translate all existing automated assembly and SISTR to analysis templates
* [x] Disable old versions of automated pipelines when pipeline is upgraded.

## Describe how you would access this feature?
Project settings page, or possibly new project page.  I think users should be able to go through the same parameter picking and launch pipeline pages for when they have things in the cart, but just instead save the analysis submission to be run when data is uploaded.

## Additional information
For first push this will only be enabled for single sample pipelines.

Issue #126, opened 2018-11-16
Clean up launch pipelines page


joshsadam:
## Describe your idea for a new feature
Launch pipelines page is getting overly large.  The `getSpecifiedPipelinePage` in particular is getting overwhelmingly large.  It should be refactored down to a more manageable size.

The JS on that page is also incredibly large and still uses Angular.  It is a very complex and heavy page.

## Describe how you would access this feature?
In PipelineController.java, generic_pipeline.html, and pipeline.launch.js

## Additional information
Main features of `getSpecifiedPipelinePage` method now are:
* Get projects in cart
* Find ref files
* Find samples & files
* Get params for workflow
* add a pile of i18n messages to page

Most of these features need to stay, but could be better separated.

`pipeline.launch.js` and `generic_pipeline.html` mostly need de-angularized

Imported from GitLab issue #567.  Originally posted on 2017/11/15 12:06PM

PR #302, opened 2019-03-04
Analysis template submission ui

tom114:
## Description of changes
Added the gui flow for adding an automated analysis submission to the project settings page.  Note the existing checkboxes are still there for assembly and sistr, they'll go away once this has been fully fleshed out.  In a later merge once everything is set to run properly, I'll add a list of the templates applied for a project.

For now, go to project settings page and click "Add Automated Pipeline".
image

That will open the pipelines page with a `?automatedProject=<pnum>` URL param, then you go through the whole pipeline launch flow as usual.  The url param will stay for all the launch pages.  The tool launch page itself is slightly modified removing the whole "Files" section and changing some messages.

image

Once you're done the buttons that show up will send you back to the project settings page.  The tools aren't actually run yet, but it'll add an `AnalysisSubmissionTemplate` to the DB.

## Related issue
Part of #240 

## Checklist

These will be done for the final merge


joshsadam:
@tom114 On the pipeline selection page, you have "Note some pipelines may be unavailable for automated analysis.", is there a way that we can only display pipelines that can do this?  I don't like allowing them to select one here and then be able to carry on when it cannot be run.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/pipelines/PipelineController.java
```diff
@@ -195,25 +178,37 @@ public String getPipelineLaunchPage(final Model model, Locale locale) {
...
-	 * @param pipelineId
-	 *            the pipeline to load
+	 *
+	 * @param model      the the model for the current request
+	 * @param principal  the user in the current request
+	 * @param locale     the locale that the user is using
+	 * @param pipelineId the pipeline to load
+	 * @param projectId  Project ID to add a {@link ca.corefacility.bioinformatics.irida.model.workflow.submission.AnalysisSubmissionTemplate}
+	 *                   if necessary (not required)
 	 * @return a page reference or redirect to load.
```
joshsadam: I know it should not be part of this merge request, but this controller is **MASSIVE**.  It is currently at 230 lines.  We need to find a better way of creating this page.

src/main/webapp/pages/pipelines/types/generic_pipeline.html
```diff
@@ -207,13 +208,19 @@ <h1 th:text="${mainTitle}"></h1>
                                         </div>
                                     </div>
                                 </div>
+                                <div class="col-sm-10" th:if="${automatedProject != null}">
```
joshsadam: Can you move this to above the parameter selection.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/pipelines/PipelineController.java
tom114: Agreed.  A lot of the analysis code has got a bit out of hand.  At the very least this could be broken down into a handful of private methods so that the main controller method is easier to read.  This is one of the refactor targets we need to look at anyway so I'd say leave it for a later full refactor merge.  See #126

src/main/webapp/pages/pipelines/types/generic_pipeline.html
tom114: This is here to replace the checkboxes for sharing results with a project.  I added the message there to show it's auto shared.  I guess I could show the project name with a checked (and disabled) checkbox instead.  What do you think?

tom114:
Removed the extra text in the "some pipelines may be unavailable..." in 00893f9

src/main/webapp/pages/pipelines/types/generic_pipeline.html
joshsadam: Sorry don't know what your comment actually means, I will come by your desk to talk to you.  This is what I was referring to:

Screen Shot 2019-03-11 at 9 47 39 AM

PR #306, opened 2019-03-07
Mysql only full group by error info

JeffreyThiessen:
Added instructions for fixing the only_full_group_by error to the FAQ page


doc/administrator/faq/index.md
```diff
@@ -162,3 +164,17 @@ collation-server = utf8_general_ci
...
+You may also run into the following error.
+
+```
+Reason: liquibase.exception.DatabaseException: Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'irida_uploader_test.ss.sample_id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
+```
+
+You can fix this by removing the `only_full_group_by` mode from sql
+
+```
+mysql -e "SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));"
```
tom114: Can we show how to set this in the mysql config file instead of in the app itself?  See the "old answer" at https://stackoverflow.com/a/37248560.  

For MariaDB and Ubuntu I believe the file would be `/etc/mysql/conf.d/mysql.cnf`

JeffreyThiessen: I'm seeing `/etc/mysql/my.cnf` as the base config file, which is what I wrote as the file to be edited in my changes. Let me know if I should change it to a more "downline" config file.

tom114:
Fix the conflicts here and this is good to go.  Thanks.

PR #307, opened 2019-03-08
Cart Galaxy Redirect and Galaxy in page status

joshsadam:
## Description of changes
galaxy-alert

 - Display an alert at the top of the page if the user in a Galaxy session
 - User can remove the Galaxy session
 - If they are in the cart and remove the galaxy session, redirect them to the pipelines page.

## Related issue
Related to #192

## Checklist


apetkau:
Thanks @josh. Looks amazing.

However, I can't seem to access any menus any more if I directly log into IRIDA. For example, if I click **Projects** nothing get's displayed:

image

I suspect the menus are opening up behind the main screen and so are hidden?

joshsadam:
Thanks @apetkau , you're right!  fixed in 244dd91a55c90f8b504d81e479b00c43c319bb72

joshsadam:
Thanks @apetkau it was a css thing, Fixed in 930683226e6763d796c82836c0bd9306d9302c55

joshsadam:
I think every you asked for (except the library name) should be updated in that last commit.  Thanks for the input @apetkau. 

@Takadonet Can you write here exactly why we don't need to have the library name input, I am passing a string that is `IRIDA-<timestamp>` as the library name.

Takadonet:
Galaxy has limitation where we cannot directly upload files into an user history so we have to first upload into a data library.  My hope that users would just share data library with others  instead of re-downloading reads again from IRIDA but it is not the case. 

Having the data library exposed in the window just added extra complexities for no benefit to the user.

apetkau:
Thanks @Takadonet and @joshsadam that makes sense. I didn't realize we were switching over to just auto-naming data libraries as `IRIDA-<timestamp>`.

apetkau:
This looks great to me @joshsadam. I'm merging it in.

Issue #347, opened 2019-04-11
Exclude xlsx analysis output files from being displayed


apetkau:
## Describe the bug

It looks like `xlsx` analysis output files are displayed in the analysis output page.

## Steps to reproduce the problem

1. Run a pipeline which produces an `xlsx` file as output (e.g., https://github.com/phac-nml/irida-plugin-amr-detection).
2. View the pipeline outputs.

## Expected behaviour

I expected the `xlsx` file to be hidden from the data displayed on the page.

## Additional context

Screenshot of output:

![image](https://user-images.githubusercontent.com/200517/55993065-c05fa980-5c73-11e9-9f58-f813ba8d312b.png)

apetkau:
Fixed in #352

PR #352, opened 2019-04-15
Excluding xlsx files from being previewed

apetkau:
## Description of changes

Excludes `.xlsx` files from being displayed in the analysis pipelines output preview.

## Related issue

Fix issue #347 

## Checklist


joshsadam:
@apetkau this looks good.  Is "black listing" all xlsx files a good thing?  I guess for previewing it is, but maybe we need to create a way to render that into the browser properly.

apetkau:
Thanks for the comments @joshsadam. We could add in a viewer for `xlsx` files. I've made an issue for this #354. But, this would require quite a bit of work to implement, so for now I think "black listing" is the best approach.

joshsadam:
Thanks @apetkau that makes complete sense.

PR #359, opened 2019-04-18
Project Setting Auto Analysis Translation

tom114:
## Description of changes
Converting the project automated assembly and SISTR settings to analysis templates.  This grabs the latest version of the pipeline and grabs all the default params for that pipeline.

Majority of the files here are test data changes.  The important parts is the liquibase XML and custom changeset updates.

## Related issue
Part of #240 

## Checklist

Checklist will be updated for final merge.


src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,168 @@
...
+
+			//insert the assembly
+			insertWorkflow(jdbcTemplate, "Automated AssemblyAnnotation", true, assemblyId, "p.assemble_uploads=1",
+					defaultAssemblyParams);
+
+		}
+
+		//update SISTR
+		if (sistrWorkflow != null) {
+			logger.debug("Upadting automated SISTR project settings");
```
apetkau: **Upadting** => **Updating**

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,168 @@
...
+		//get the workflow information from the service
+		try {
+			sistrWorkflow = workflowsService.getDefaultWorkflowByType(BuiltInAnalysisTypes.SISTR_TYPING);
+		} catch (IridaWorkflowNotFoundException e) {
+			logger.warn("SISTR workflow not found.  Automated SISTR will not be converted to analysis templates.");
+			//Note this will definitely happen in the galaxy CI tests as only SNVPhyl and a test workflow are configured.
+		}
+
+		//update assemblies
+		if (assemblyWorkflow != null) {
```
apetkau: It's not really necessary, but you could move the below code inside the `try-catch` block and it saves you having to check for `null` here.

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	private void insertWorkflow(JdbcTemplate jdbcTemplate, String name, boolean updateSamples, UUID workflowId,
+			String where, List<IridaWorkflowParameter> params) {
+		/*
+		 * we're borrowing the 'automated' flag here to mark entries we need to add params to later.  at this point
+		 * there'll be nothing with a '1' in the automated flag.  we'll clear it later.
+		 */
+
+		int updateSampleBit = updateSamples ? 1 : 0;
```
apetkau: Would a `boolean` work here, since that's what the `updateSampleBit` is representing? Not too sure about if it needs to be `0` or `1` or `false` or `true` here.

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,168 @@
...
+								+ "', '" + p.getDefaultValue()
+								+ "' FROM analysis_submission a where a.name=? AND a.automated=1";
+
+				jdbcTemplate.update(assemblyParamInsert, name);
+			}
+
+			// remove the automated=1
+			String removeAutomatedSql = "UPDATE analysis_submission SET automated=null WHERE DTYPE = 'AnalysisSubmissionTemplate' AND name=?";
+			jdbcTemplate.update(removeAutomatedSql, name);
+		}
```
apetkau: Could you log the case where nothing gets added, which could help with debugging later?

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,168 @@
...
+
+		int update = jdbcTemplate.update(assemblyInsert);
+
+		//if we added anything, add the params
+		if (update > 0) {
+
+			// Insert the default params for the analysis type
+			for (IridaWorkflowParameter p : params) {
+				String assemblyParamInsert =
+						"INSERT INTO analysis_submission_parameters (id, name, value) SELECT a.id, '" + p.getName()
```
apetkau: I believe the parameter name/values be escaped when inserting into the database. It could be possible they contain quotes and other special characters (I don't think they do, but just in case).

src/main/resources/ca/corefacility/bioinformatics/irida/database/changesets/19.04/analysis-templates.xml
```diff
@@ -24,26 +24,39 @@
             <column name="automated" type="bit(1)"/>
         </addColumn>
 
-        <!--updating old automated runs-->
-        <sql>UPDATE analysis_submission SET automated=0</sql>
+        <!--Drop old notnulls since the table now has the template also encoded -->
```
apetkau: Could you move this entire file to the database changesets for the latest release?

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
tom114: Logging added in 0b67979ce1

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
tom114: This was done just since a `bit(1)` type is a `1` or `0`, but yeah mysql is smart enough to translate a `true` or `false` in the query.  Removed this in 0b67979ce1

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
tom114: Sure it cleans it up a bit.  Done in 0b67979ce1

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
tom114: Fixed in 0b67979ce1

src/main/resources/ca/corefacility/bioinformatics/irida/database/changesets/19.04/analysis-templates.xml
tom114: Merged development up and moved to 19.09 in da81690

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
tom114: Changed all the insertions to be paramaterized in 4d037c6.  It involved a couple more `SELECT` queries, but this will be a quick changeset so no big deal.

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,185 @@
...
+
+				//build the argument list for the query
+				List<Object[]> submissionParamArgs = submissionIds.stream()
+						.map(i -> {
+							return new Object[] { i, p.getName(), p.getDefaultValue() };
+						})
+						.collect(Collectors.toList());
+
+				//then insert the params for each submission
+				String paramInsert = "INSERT INTO analysis_submission_parameters (id, name,value) VALUES (?, ?, ?)";
```
apetkau: Same comment here. There's no audit tables being created for `analysis_submission_parameters`.

src/main/java/ca/corefacility/bioinformatics/irida/database/changesets/AutomatedAnalysisToTemplate.java
```diff
@@ -0,0 +1,185 @@
...
+
+		//build the params for the insert submission
+		List<Object[]> queryParams = projectIds.stream()
+				.map(p -> {
+					return new Object[] { name, updateSamples, workflowId.toString(), p };
+				})
+				.collect(Collectors.toList());
+
+		//then insert the submisisons for each project
+		String submissionInsert = "INSERT INTO analysis_submission (DTYPE, name, created_date, priority, update_samples, workflow_id, submitter, submitted_project_id, automated) VALUES ('AnalysisSubmissionTemplate', ?, now(), 'LOW', ?, ?, 1, ?, 1)";
```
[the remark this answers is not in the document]
tom114: In the past our stance has generally been to **not** add rows to the audit tables in cases that are complicated such as this one.  This is because it's quite hard to retroactively fully replicate what Envers does when it's creating revisions.  We would need to add `Revisions` entries, users, clients, timestamps, and most difficulty attaching the new `_AUD` entry to the *right* revision since it may have been created along with a project, etc.  With all that it would be easy to end up with the revisions and `_AUD`s in an inconsistent (or incorrect) state.

Given that stance, if you need to see when something like that was created, you can do it in the database backups.  The AUD records for that are still alive and well there.

Something we could do for posterity is add something to the `analysisDescription` field of the submission saying it was "translated from an old auto assembly on this date" etc. That way at least people can see where it came from.  How do you feel about that?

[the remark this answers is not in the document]
apetkau: I think that makes sense. I can see it being quite complicated to replicate everything that the audit tables would have entered. I think adding that message in the `analysisDescription` field is a good compromise.

Something else I was concerned about was whether or not new audit entries would still work if old ones weren't created? But, I think each entry is independent from the previous ones, so it won't make much of a difference.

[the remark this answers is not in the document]
tom114: Added a description in 27ca2d17fd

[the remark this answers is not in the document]
apetkau: This works great. Thanks.

apetkau:
This looks great Tom. Thanks.

PR #375, opened 2019-05-01
Renamed release to 19.05

tom114:
## Description of changes
Renamed release descriptions to 19.05 from 19.04.  We missed the 19.04 window due to a couple bugs created in this release's development.

## Related issue
N/A

## Checklist


doc/administrator/upgrades/index.md
```diff
@@ -11,7 +11,7 @@ description: "Upgrade Notes"
 
 The majority of IRIDA's upgrade notes can be seen at <https://github.com/phac-nml/irida/blob/master/UPGRADING.md>.  When there are more major upgrades to the IRIDA system, database, deployment, etc. this page will go further in depth about how to properly back up your system, perform the upgrade, and recover from problems.
 
-# 19.04
+# 19.05
```
apetkau: Since some people have already upgraded to `19.04-SNAPSHOT` I'm wondering if we can add a section in here listing the custom changes they will have to make?

Specifically, they will have to log into the database **before** the upgrade and run the following SQL command:

```sql
UPDATE DATABASECHANGELOG SET FILENAME = REPLACE(FILENAME,'classpath:ca/corefacility/bioinformatics/irida/database/changesets/19.04','classpath:ca/corefacility/bioinformatics/irida/database/changesets/19.05');
```

This will make the same changes we are making to the database change log files (from 19.04 to 19.05) in their database.

If they do not run these changes, then when they attempt to install IRIDA `19.05` they will see the following errors in the database:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'springLiquibase' defined in class ca.corefacility.bioinformatics.irida.config.data.IridaApiJdbcDataSourceConfig: Invocation of init method failed; nested exception is liquibase.exception.
MigrationFailedException: Migration failed for change set classpath:ca/corefacility/bioinformatics/irida/database/changesets/19.05/add-email-pipeline-result-to-analysis-submission.xml::email-pipeline-result::deep:
     Reason: liquibase.exception.DatabaseException: Duplicate column name 'email_pipeline_result' [Failed SQL: ALTER TABLE irida_test.analysis_submission ADD email_pipeline_result BIT(1) DEFAULT 0 NOT NULL]
...
```

We should also mention that it is strongly recommended to not upgrade to a development version of IRIDA on production systems as this could lead to possibly lead to things breaking if we refactor some code before the official release (as we are doing now).

doc/administrator/upgrades/index.md
```diff
@@ -11,7 +11,7 @@ description: "Upgrade Notes"
...
 ## FastQC translation to filesystem
```
apetkau: I wonder if we should also include in the **In case of error** section of this document both that you will likely be unable to access IRIDA in case of an error, and if you try to re-run Tomcat again you may encounter messages like below and will have to search previous messages in the log file to track down the first database upgrade error?

```
Change Set classpath:ca/corefacility/bioinform
atics/irida/database/changesets/19.05/fastqc-filesystem.xml::fastqc-filesystem::tom failed.  Error: Duplicate column name 'analysis_id' [Failed SQL: ALTER TABLE irida_test.analysis_output_file ADD analysis_id BIGINT NULL]
```

This could be added as another pull request (since it's unrelated to switching the version).

doc/administrator/upgrades/index.md
tom114: 2 thoughts to solve this:
1. If someone upgraded to a development version we can't support production upgrades to the database.  I think it's fair to add the `strongly recommended to not upgrade to a development version...` thing but to me it should be no surprise if they can't upgrade from a dev to prod.
2. We could just leave the db directory as `19.04`. Maybe add a .txt file explaning what went on for anyone who's really curious, but it wouldn't cause us any issues.

tom114:
Updated pom.xml in 8ae417f

doc/administrator/upgrades/index.md
apetkau: Thanks. Yes, I was wondering if we should just leave the db directory to `19.04`. Maybe that solution is better after all and things should just work without running custom SQL commands.

For strongly recommended, maybe we should even go to the point of saying "we won't support upgrades from development to production though we will try to help out if requested". As we can't guarantee really much of anything from development to production.

tom114: Rolled the db version changes back to `19.04` in fce336f

apetkau: Looks good. Thanks.

doc/administrator/upgrades/index.md
tom114: I'd say lets do this in a later PR because it will apply to more than just this upgrade.  It along with the "not supported if you use development versions of IRIDA" thing need to be linked from lots of places.

Also more directly on this comment, I think it's pretty clear that you wouldn't be able to run IRIDA in case of a DB upgrade issue.  We have lots of mitigation information for this particular update.

apetkau: Sounds good :+1:

apetkau:
Look good. Thanks.

PR #399, opened 2019-06-03
Fix/galaxy fastqsanger

### LINKED ISSUE #357: Gzipped fastq files uploaded to Galaxy as uncompressed type
## Describe the bug

As part of an effort to support `.fastq.gz` files (#355) I noticed that gzipped fastq files get uploaded to Galaxy as type **fastqsanger** (instead of type **fastqsanger.gz**).

Most pipelines still work in this case, but one of them (biohansel) does not. We should fix the code so IRIDA sets the correct type in Galaxy.

The code is set in https://github.com/phac-nml/irida/blob/75c784adae046ae765cc0eb8f32b8c163923193a/src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/AnalysisCollectionServiceGalaxy.java#L87

1. Configure IRIDA to not decompress files `file.processing.decompress=false`.
2. Upload `fastq.gz` files and send for data analysis.
3. Check file type in Galaxy.

## Expected behaviour

I expected IRIDA to set the proper file type in Galaxy (**fastqsanger** or **fastqsanger.gz**) depending on the files being sent to Galaxy.

### Galaxy `fastqsanger` type

image

### Galaxy `fastqsanger.gz` type

image

apetkau:
## Description of changes

When running a pipeline, properly sets the datatype of files sent to Galaxy to either **fastqsanger** or **fastqsanger.gz** depending on if the files are gzipped. For example:

image

## Related issue

Fixes #357 

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/model/irida/IridaSequenceFile.java
```diff
@@ -25,6 +25,13 @@
 	 * @return the physical location of the file.
 	 */
 	public Path getFile();
+	
+	/**
+	 * Returns whether or not the file is gzipped.
+	 * 
+	 * @return True if gzipped, false otherwise.
+	 */
+	public boolean isGzipped();
```
tom114: This class (Along with all the Irida* model interfaces) really aught to be deprecated and removed.  Was the Galaxy stuff using `IridaSequenceFile` already?  If so we can just remove it in a later merge.

tom114: Added an issue for removing them #405

run-tests.sh
```diff
@@ -131,7 +131,7 @@ test_galaxy_internal() {
 	shift
 
 	docker run -d -p $GALAXY_PORT:80 --name $GALAXY_DOCKER_NAME -v $TMP_DIRECTORY:$TMP_DIRECTORY -v $SCRIPT_DIR:$SCRIPT_DIR $GALAXY_DOCKER && \
-	mvn clean verify -B -P$profile -Djdbc.url=$JDBC_URL -Dirida.it.rootdirectory=$TMP_DIRECTORY -Dtest.galaxy.url=$GALAXY_URL -Dtest.galaxy.invalid.url=$GALAXY_INVALID_URL -Dtest.galaxy.invalid.url2=$GALAXY_INVALID_URL2 -Dsequence.file.base.directory=$SEQUENCE_FILE_DIR -Dreference.file.base.directory=$REFERENCE_FILE_DIR -Doutput.file.base.directory=$OUTPUT_FILE_DIR -Djdbc.pool.maxWait=$DB_MAX_WAIT_MILLIS $@
+	mvn clean verify -B -P$profile -Dfile.processing.decompress=false -Djdbc.url=$JDBC_URL -Dirida.it.rootdirectory=$TMP_DIRECTORY -Dtest.galaxy.url=$GALAXY_URL -Dtest.galaxy.invalid.url=$GALAXY_INVALID_URL -Dtest.galaxy.invalid.url2=$GALAXY_INVALID_URL2 -Dsequence.file.base.directory=$SEQUENCE_FILE_DIR -Dreference.file.base.directory=$REFERENCE_FILE_DIR -Doutput.file.base.directory=$OUTPUT_FILE_DIR -Djdbc.pool.maxWait=$DB_MAX_WAIT_MILLIS $@
```
tom114: Any issue leaving them compressed here and us not testing it uncompressed?  Or will similar functions get tested uncompressed in the pipeline testing?

src/main/java/ca/corefacility/bioinformatics/irida/model/sequenceFile/SequenceFile.java
```diff
@@ -123,6 +123,16 @@ public SequenceFile(Path sampleFile) {
 		this();
 		this.file = sampleFile;
 	}
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public boolean isGzipped() {
+		String fileExtension = com.google.common.io.Files.getFileExtension(this.file.getFileName().toString());
```
tom114: I was looking at this sort of thing instead of the file extension, but I'm not sure if it's worth the extra effort.  https://stackoverflow.com/questions/30507653/how-to-check-whether-file-is-gzip-or-not-in-java
 Particularly since this requires the file to be read and it may take longer than it's worth.  In general I think the file extension checking is fine but I wanted to discuss whether it's worth it.

src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/SequenceFilePathType.java
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * Creates a new {@link SequenceFilePathType}.
+	 * 
+	 * @param sequenceFile The specific sequence file.
+	 */
+	public SequenceFilePathType(IridaSequenceFile sequenceFile) {
+		checkNotNull(sequenceFile, "sequenceFile is null");
+
+		this.path = sequenceFile.getFile();
+		this.fileType = sequenceFile.isGzipped() ? InputFileType.FASTQ_SANGER_GZ : InputFileType.FASTQ_SANGER;
```
tom114: Think this is something we could just add to `SequenceFile` or `SequencingObject`?  It wouldn't need to be a property in the database, it could just be a method.  That way we can ditch the extra class too.

src/main/java/ca/corefacility/bioinformatics/irida/model/irida/IridaSequenceFile.java
apetkau: Yes, the Galaxy code was using `IridaSequenceFile`. Thanks for making another issue.

src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/SequenceFilePathType.java
apetkau: The problem with adding to `SequenceFile` is that this is Galaxy-specific code (that is, `InputFileType` is specific to Galaxy). I did not want to mix Galaxy code with the IRIDA objects, so I kept the code separate.

tom114: Understood, but this feels like an unnecessary wrapper class that could maybe be solved with just a static method call.  Previously I see you were passing around the `Path` instead of this `SequenceFilePathType`.  Couldn't you just continue sending around the `Path` and then check whether the file is zipped when you need it?  The only place I see the `InputFileType` is at https://github.com/phac-nml/irida/pull/399/files#diff-3f2e082b327e44d9660a84b42fb07165R202.  That could reduce this whole changeset down to just a method call.  Just replicate the functionality from `public boolean isGzipped()` in `SequenceFile`.

Thoughts on that?

run-tests.sh
apetkau: I've changed this behaviour in 64d721c474f2e542e31238caaa925134a7fd341a, 93a2b2a72651d56eeddf7902518a37f00f7183e3, and 6078af440582bfd55f11533acefed1df8ac5487f so that `file.processing.decompress=true` is always set.

In 93a2b2a72651d56eeddf7902518a37f00f7183e3 specifically I modify the `GzipFileProcessor` so that it can be dynamically enabled and disabled. That way, for specific tests, I can disable it, but it should always default to enabled.

src/main/java/ca/corefacility/bioinformatics/irida/model/sequenceFile/SequenceFile.java
apetkau: We actually already implemented this in the `GzipFileProcessor`, where we were checking for file magic numbers to see if it was gzipped. I moved this code to a `FileUtils` class, and am using this same code (instead of file extension) to check if the file is gzipped. Done in 051f2923711b2ad78e6408fe3ec6e182e0b70c8a

src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/SequenceFilePathType.java
apetkau: That makes sense. I've removed `SequenceFilePathType` in 6ed7026f0b624686040635d38893c171aab90801 and implemented what you mentioned.

apetkau:
@tom114 I believe I've addressed all your issues and I tested this out again and it all works with both compressed and uncompressed data. It should be good to review again.

src/main/java/ca/corefacility/bioinformatics/irida/pipeline/upload/galaxy/GalaxyLibrariesService.java
```diff
@@ -254,6 +255,12 @@ public Void call() throws Exception {
 
 		return datasetLibraryIdsMap;
 	}
+	
+	private InputFileType getFileType(Path path) throws IOException {
```
tom114: Add comments for this method please.

run-tests.sh
```diff
@@ -97,19 +97,19 @@ exit_error() {
 }
 
 test_service() {
-	mvn clean verify -B -Pservice_testing -Djdbc.url=$JDBC_URL -Dirida.it.rootdirectory=$TMP_DIRECTORY -Dsequence.file.base.directory=$SEQUENCE_FILE_DIR -Dreference.file.base.directory=$REFERENCE_FILE_DIR -Doutput.file.base.directory=$OUTPUT_FILE_DIR -Djdbc.pool.maxWait=$DB_MAX_WAIT_MILLIS $@
+	mvn clean verify -B -Pservice_testing -Djdbc.url=$JDBC_URL -Dfile.processing.decompress=true -Dirida.it.rootdirectory=$TMP_DIRECTORY -Dsequence.file.base.directory=$SEQUENCE_FILE_DIR -Dreference.file.base.directory=$REFERENCE_FILE_DIR -Doutput.file.base.directory=$OUTPUT_FILE_DIR -Djdbc.pool.maxWait=$DB_MAX_WAIT_MILLIS $@
```
tom114: Is the default `file.processing.decompress=true`?  If so does this need set for all these tests?

apetkau: Yes, that is the default. The reason I'm setting it explicitly here is if you have set `file..processing.decompress=false` in your `/etc/irida/irida.conf` file, and then run the tests on your own machine (like I sometimes do), they won't work properly.

tom114: Ah understood.  Yeah that's good thanks.

src/main/java/ca/corefacility/bioinformatics/irida/pipeline/upload/galaxy/GalaxyLibrariesService.java
apetkau: Added in f738abcb49

PR #417, opened 2019-06-25
Projects Page Refactor

joshsadam:
## Description of changes
The goal of this merge request is to start removing the old style of DataTables from IRIDA.  Here we are replaces the listing of projects DataTables with the default table from ant.design.  

## Related issue
N/A

## Checklist

* [X] User documentation updated for UI or technical changes.


src/main/webapp/resources/js/pages/projects/list/ProjectsTable.jsx
```diff
@@ -0,0 +1,224 @@
...
+      }
+    });
+  };
+
+  const onSearch = value =>
+    dispatch({
+      type: TYPES.SEARCH,
+      payload: {
+        search: value,
+        currnet: 0
```
deepsidhu85: Typo here. Should be `current: 0`

joshsadam: Fixed in abd3b66b7d8f7829b2a4ec652971d8ab470c8774.  It actually needs to be `current: 1`, since the UI table uses 1 as the initial page, this is corrected for when the server call is actually made.

PR #400, opened 2019-06-03
webpack to handle internationalizations

joshsadam:
## Description of changes
For each webpack entry point, webpack will scan for `i18n` method calls and pull the translation from the Spring message properties files.

## Related issue
Link to the GitHub issue this pull request addresses using the `#issuenum` format.  If it completes an issue, use `Fixes #issuenum` to automatically close the issue.

## Checklist
Things for the developer to confirm they've done before the PR should be accepted:

* [x] CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change.
* [X] Tests added (or description of how to test) for any new features.
* [X] User documentation updated for UI or technical changes.

PR #418, opened 2019-07-02
Spring using user's i18n settings

tom114:
## Description of changes
Made i18n actually read the files, properly fallback to default language, and have a language selector on the user edit/creation page.  Right now the lang selector is locked to `English` since no full internationalizations have been done, but it could be easily extended.

## Related issue
No related issue, but related to PR #400 

## Checklist


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Changes
 * [UI/Developer]: Removed `datatables` as a bower dependency.
 * [UI/Developer]: Removed unused dependencies from `bower` (`angular-bootstrap-switch`, `animate.css`, `angular-datatables`, `angular-daterangepicker-enhanced`, `lodash`, `clipboard`, and `jszip`)
 * [UI]: Removed lighthouse modals from the sequence files page.  Loading full images.  Removed dependency `angular-bootstrap-lightbox` from `bower`.
+* [Developer]: Internationalization faetures of Spring now being used properly.  An interceptor was previously set up, but wasn't being taken advantage of in the UI.  Currently only English is available, but could be added by developers as needed.
```
joshsadam: Typo `faetures `

src/main/java/ca/corefacility/bioinformatics/irida/config/services/IridaApiServicesConfig.java
```diff
@@ -176,7 +157,7 @@ public MessageSource messageSource() {
 			source.setBasenames(RESOURCE_LOCATIONS);
 		}
 
-		source.setFallbackToSystemLocale(false);
+		source.setFallbackToSystemLocale(true);
```
joshsadam: What happens if the system locale is not a valid language (if this is possible)?

tom114: To go one step further, it might be an issue if the system locale is a **valid** non_en locale as it wouldn't fall back to  `en`.  I think the right thing to do is to set this back to `false`, then rename `messages_en.properties` to `messages.properties`.  That makes it the global default (see https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/AbstractResourceBasedMessageSource.html#setFallbackToSystemLocale-boolean-). 

If we had full `messages_x.properties` for all (reasonable) languages the system locale would make sense, but since the other languages will likely be partial files I think this makes sense.  Thoughts?

joshsadam: I think that is a really good idea.

tom114: Done in 11b9b5d

CHANGELOG.md
tom114: Fixed in 
c63b297

tom114:
Changed this to have the available locales be configurable in 3a09987

joshsadam:
This is actually breaking the cart page when attempting to access it as a language that does not have the key it is looking for:

```
org.springframework.context.NoSuchMessageException: No message found under code 'workflow.BIO_HANSEL.title' for locale 'fr'.
	at org.springframework.context.support.AbstractMessageSource.getMessage(AbstractMessageSource.java:159)
	at ca.corefacility.bioinformatics.irida.ria.web.pipelines.PipelineController.createPipeline(PipelineController.java:565)
	at ca.corefacility.bioinformatics.irida.ria.web.pipelines.PipelineController.getWorkflowTypes(PipelineController.java:542)
```

tom114:
Switched back to falling back to system language in b8364b472a.  The problem is that pipelines (and plugin pipelines) still named their files as `messages_en.properties` so by removing the system fallback they were finding no messages.  Putting the system fallback in should be fine since we're setting the default locale in https://github.com/phac-nml/irida/blob/development/src/main/java/ca/corefacility/bioinformatics/irida/config/web/IridaUIWebConfig.java#L120.  I've confirmed that switching this will change the default "system" locale that's used for the messagesource, so all the pipelines will still be ok.  You can test it out though.

src/main/java/ca/corefacility/bioinformatics/irida/config/services/IridaApiServicesConfig.java
```diff
@@ -418,5 +428,20 @@ public SpringTemplateEngine exportUploadTemplateEngine() {
 	public NreplServerSpring nRepl() {
```
joshsadam: I just see this here and it is not related, but I thought we got rid of the REPL stuff?

Issue #451, opened 2019-08-15
Uploading to NCBI via FTP is not explicitly being set to "PASSIVE" mode


apetkau:
## Describe the bug
We use FTP to upload sequence reads to NCBI. There are two modes available in FTP to handle data connections:

1. **Active mode**: The client sends a `PORT` command telling the server (NCBI) which port it should connect to on the client's end to transfer data. This mode only works if the client is not behind a firewall/NAT (e.g., it has a public IP address and all ports are open to the Internet).
2. **Passive mode**: The client sends a `PASV` command telling the server (NCBI) it wants to transfer data in passive mode. The FTP server then responds to the client with a particular (random) port to connect to for transferring data. This mode works even when the client is behind a firewall/NAT.

According to the documentation, the default is active mode:

>Default FTPClient constructor. Creates a new FTPClient instance with the data connection mode set to ACTIVE_LOCAL_DATA_CONNECTION_MODE , the file type set to FTP.ASCII_FILE_TYPE , the file format set to FTP.NON_PRINT_TEXT_FORMAT , the file structure set to FTP.FILE_STRUCTURE , and the transfer mode set to FTP.STREAM_TRANSFER_MODE - https://commons.apache.org/proper/commons-net/apidocs/org/apache/commons/net/ftp/FTPClient.html#FTPClient()

This would never work for us (and likely most other people). I do not know how this was even working to begin with? This default setting must also be controlled elsewhere (maybe in system configuration?).

## Steps to reproduce the problem
1. I installed IRIDA from our virtual machine and configured our NCBI FTP account.
2. I tried uploading test files. The connection would freeze up. I would see the following error as a response from the server:

    ```
    500 Illegal PORT command
    ```

    This suggests that our FTP client is trying to send files in active mode (issuing a PORT command instead of PASV), and the server is responding that this is invalid (maybe configured only for passive mode).

I'm still not sure how, on some setups, we can upload files to NCBI though without that error.

## Expected behaviour
I expected upload to NCBI to work and just default to passive mode.

## Additional context
N/A

apetkau:
Fixed in #452

PR #452, opened 2019-08-15
Hotfix/ftp mode

apetkau:
## Description of changes
Adds the ability to switch between active/passive FTP mode through a config setting `ncbi.upload.ftp.passive=true`.

In addition to fixing the below issue, this appears to fix the slow upload speed I could never figure out.

Before, I was getting uploads from my machine to NCBI between 3-4 MB/s, but using regular FTP on the command-line I could get up to 40 MB/s.

After this fix, when defaulting to passive mode, I get uploads up to 40 MB/s (so maxing out the connection to NCBI).

I have no idea why this is right now, but it's a nice benefit.

If you wish to test, you'll have to configure your local IRIDA instance to connect to NCBI (loading data in the `/submit/Test` directory instead of `/submit/Production`).

## Related issue
Fixes issue #451 

## Checklist


apetkau:
@tom114 switched to merging to `development`.

Note, I suspect the two javascript files that are changed were due to the automated reformatting that gets run with `git commit`, since I didn't touch any javascript files.

PR #460, opened 2019-09-13
Update Sequencing Runs to use the ant.design table

joshsadam:
## Description of changes
Update the sequence run listing page to use ant.design tables.

## Related issue
#413 

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/TableResponse.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+
+/**
+ * Response sent when items are requested for a table.
+ */
+public class TableResponse {
+	private List<TableModel> dataSource;
+	private Long total;
+
+	public TableResponse(List<TableModel> analyses, Long total) {
```
deepsidhu85: Since the responses for tables has been generalized we should probably change `analyses` here to `dataSource`

src/main/webapp/resources/js/contexts/PagedTableContext.js
```diff
@@ -37,16 +41,21 @@ function AnalysesProvider({ children }) {
...
-      current: tableState.current - 1,
-      pageSize: tableState.pageSize,
-      sortColumn: tableState.column,
-      sortDirection: tableState.order,
-      search: tableState.search,
-      filters: tableState.filters
-    }).then(({ analyses, total }) => {
-      setTableState({ ...tableState, ...{ total, analyses, loading: false } });
-    });
+    axios
```
deepsidhu85: Would it not be better to have this axios call in a js file within the js/apis folder?

joshsadam: Good idea, updated in 369ac9a0b60829c75e811f8084b31ef0b929c623

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/TableResponse.java
joshsadam: Thanks, renamed in 10a490cebdb66f6a8be898a4c352097e5d81ab1c

PR #482, opened 2019-10-09
Concatenate zipped fastq

### LINKED ISSUE #481: Sequence concatenator doesn't work with compressed data
## Describe the bug
When concatenating a `.fastq.gz` sequence file from the UI, the result is a `.fastq` file that doesn't process correctly.

1. On sample/files page click concatenate.
2. Select `.fastq.gz` files to concatenate and click go.
3. Result is `.fastq` file that fails processing.

## Expected behaviour
Files to be correctly concatenated and valid `.fastq.gz` files.

## Additional context
File concatenator is assuming files are `.fastq` extensions and not correctly merging the zipped files.

tom114:
## Description of changes
Fixed the file concatenation methods to work with `.fastq.gz` data.  Previously the file extension was hardcoded as `.fastq` so the file processing would fail.  Also the method we were using to concatenate the files didn't work with gzipped data.

To test, upload multiple files to a sample, wait for them to file process, then concatenate from the sample/files page.  Then change the `file.processing.decompress=` value in your `/etc/irida/irida.conf` file and try the same again.

## Related issue
Fixes #481 

## Checklist

### DIFF
--- a/src/main/java/ca/corefacility/bioinformatics/irida/processing/concatenate/SequencingObjectConcatenator.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/processing/concatenate/SequencingObjectConcatenator.java
@@ -19,6 +19,7 @@ import java.util.Optional;
  * @param <Type> the {@link SequencingObject} class to concatenate
  */
 public abstract class SequencingObjectConcatenator<Type extends SequencingObject> {
+	//Valid extensions to try to concatenate with this tool
 	private static final List<String> VALID_EXTENSIONS = Lists.newArrayList("fastq", "fastq.gz");
 
 	/**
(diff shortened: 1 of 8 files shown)


src/main/java/ca/corefacility/bioinformatics/irida/processing/concatenate/SequencingObjectConcatenator.java
```diff
@@ -1,63 +1,97 @@
...
+	 * @throws ConcatenateException if there is an error appending the file
 	 */
 	protected void appendToFile(Path target, SequenceFile file) throws ConcatenateException {
 
-		try (FileWriter fw = new FileWriter(target.toFile(), true);
-				BufferedWriter writer = new BufferedWriter(fw);
-				FileReader reader = new FileReader(file.getFile().toFile())) {
-
-			IOUtils.copy(reader, writer);
+		try (FileChannel out = FileChannel.open(target, StandardOpenOption.CREATE, StandardOpenOption.APPEND,
```
tom114: Switched this to use `FileChannel` instead of the `FileWriter`.  `FileWriter` was having issues reading/writing the gzip files.  `FileChannel` is also a fair bit faster.

tom114:
Re: The zipping after change in `file.processing.decompress`, I think that's more of an option in your initial setup.  Don't think we need to address that yet.

PR #483, opened 2019-10-09
Fix client edit form to correctly populate authorized grant types

ericenns:
## Description of changes
Fixes issue where dropdown in clients edit form doesn't select "password" or "authorization_code" when refresh token is enabled. If refresh token is enabled it always selects "password" regardless. 

## Related issue

## Checklist


ericenns:
> Thanks that works well. Can I ask for a small addition? Can you add the refresh token validity for if refresh token is checked off?
> 
> Also add an entry to the changelog about this change.

Fixed in latest commit, also found issue with toggling of hidden fields in both client and edit forms which is fixed.

PR #495, opened 2019-10-18
User role reset

### LINKED ISSUE #114: Create User defaults Role to User after error in password
When creating a new user, if Require User Activation is unchecked and you mistype the second password field, when the page refreshes, all fields retain values except the role switches back to User.

Imported from GitLab issue #548.  Originally posted on 2017/11/02 11:24AM\Posted by Jen Cabral

tom114:
## Description of changes
Added fix to user creation page to set the user's role to the previously selected value if the server-side validation fails.  This is hard to duplicate or test since all the fields are checked on client-side validation, but in the odd case that server validation fails, it should now add the user's role.

## Related issue
Fixes #114 

## Checklist


tom114:
Changelog conflict resolved.

PR #577, opened 2020-01-22
Update to Ant Design v4

joshsadam:
## Description of changes
Update to the next version of [Ant Design](https://next.ant.design/) (v4).  The main difference is an incredibly smaller file size do to a change in how the icons are used.

One major issue encountered was the change to how forms are created, the process is much smoother.  This appears to only effect one page currently - saving a linelist template.  It is fixed in the PR, but will get refactored again shortly when working on default metadata templates.  

## Related issue
N/A

## Checklist


src/main/webapp/resources/js/contexts/PagedTableContext.js
```diff
@@ -66,14 +66,20 @@ function PagedTableProvider({ children, url }) {
   const handleTableChange = (pagination, filters, sorter) => {
     const { pageSize, current } = pagination;
     const { order, field } = sorter;
+
+    // Need to remove null filters in ant design v4
```
deepsidhu85: Is this used by say for example "States" is null then it won't display the filter icon for States?

joshsadam: No ant v4 just sends null (previously nothing) which messes with our server code.  So if they are null we remove them client side before sending to server.

src/main/webapp/resources/js/pages/projects/linelist/components/Export/ExportDropdown.jsx
```diff
@@ -1,7 +1,11 @@
...
-
 import PropTypes from "prop-types";
 import { Button, Dropdown, Menu } from "antd";
+import {
+  DownloadOutlined,
+  FileExcelOutlined,
+  FileOutlined
+} from "@ant-design/icons";
 
 export function ExportDropDown(props) {
```
deepsidhu85: This export dropdown button doesn't provide the user with the context that it is an actual dropdown. Maybe the download icon within can be replaced with arrow icons (facing up when button is hovered and facing down otherwise)?

PR #606, opened 2020-02-20
Changed trim for tablerequests

### LINKED ISSUE #605: ClientsControllerTest failing after hotfix
## Describe the bug
`ClientsControllerTest.testGetAjaxClientList` is failing due to the hotfix lower casing the search terms coming in to the new ant tables.

1. ran `mvn test`
2. `ClientsControllerTest.testGetAjaxClientList` failed
3. 

## Expected behaviour
`ClientsControllerTest.testGetAjaxClientList` pass

## Additional context
It's a null pointer due to the change to ant design on the clients table.  It looks like no search term is being sent to the tablerequest.

tom114:
## Description of changes
Changed the search term to be trimmed when it's set for the `TableRequest` instead of when it's requested.  The issue was that if there's no search term and the getter is called, it's trying to trim it then which would throw a nullpointer.

## Related issue
Fixes #605 

## Checklist

~* [] CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change.~ Not bothering because we caused this error with the hotfix.  Not really a change.

### DIFF
--- a/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/tables/TableRequest.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/tables/TableRequest.java
@@ -44,7 +44,11 @@ public class TableRequest {
 	}
 
 	public void setSearch(String search) {
-		this.search = search.trim();
+		if (search != null) {
+			this.search = search.trim();
+		} else {
+			search = null;
+		}
 	}
 
 	/**
--- a/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/tables/TableRequest.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/tables/TableRequest.java
@@ -40,6 +40,9 @@ public class TableRequest {
 	}
 
 	public String getSearch() {
+		if (search == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/ca/corefacility/bioinformatics/irida/ria/web/models/tables/TableRequest.java
```diff
@@ -40,11 +40,11 @@ public void setSortColumn(String sortColumn) {
 	}
 
 	public String getSearch() {
-		return search.trim();
+		return search;
 	}
 
 	public void setSearch(String search) {
-		this.search = search;
+		this.search = search.trim();
```
joshsadam: Wouldn't this still through an error if the search argument is `null`?

tom114: I don't think it should happen that a null ever gets set here, but I did it just in case in 90b5c05.

joshsadam: Thanks, that looks good.  Agreed thought that it shouldn't but safe is good.

PR #631, opened 2020-03-17
Remove global angular ui bootstrap

joshsadam:
## Description of changes
Currently `angular-ui-bootstrap` is loaded in `app.js` which causes it to be loaded on every page load.  As this is mostly unused now, we only want to load it on the pages that have it as a direct dependency.

## Related issue
N/A

## Checklist


joshsadam:
@deepsidhu85 added missing i18n string in d8b9dfbec3029340890182ba235fda537aeb0549
I would assume that is why it crashes.

PR #661, opened 2020-04-21
Rest project sample add graceful

### LINKED ISSUE #593: Sharing samples via REST API: Sharing is only applied partially if at least one sample already exists in target project.
## Describe the bug
Briefly describe the issue you encountered.

When sharing a set of samples to another project via the REST API, sharing is only applied partially if at least one sample already exists in the target project.

1. Create Project 1 that includes samples with IDs 1,2,3
2. Create Project 2 and share sample 2 from Project 1 to Project 2 (via the web interface)
3. Attempt to share samples 1,2,3 to Project 2 via the REST API (`POST` `[1,2,3]` to `/projects/2/samples`)

## Expected behaviour
What did you expect to happen?

I'd expect samples 1 and 3 to both be shared to Project 2, but only sample 1 is actually shared.

## Additional context
Additional information to help solve the problem.  Test files, screenshots, IRIDA version information can go here.

tom114:
## Description of changes
Changed the way the REST API responds when copying samples to projects if they already exist.  Previously the REST API would add "some" of the samples in the list, then when it ran into a duplicate it would throw an exception and return.

Ex: if adding samples [1,2,3] and sample 2 already existed, it would add 1, fail adding 2, but then nothing would happen to 3.

The new functionality here is to warn on the server side if the sample already exists, but carry on as normal as if it was successful.  In the above sample this would result in samples 1 & 3 being added to the project, and sample 2 already existed there.

## Related issue
Fixes #593 

## Checklist


JeffreyThiessen (approved):
I testing it and it looks good to go. 👍

PR #687, opened 2020-05-12
Create new user group using Ant Design

joshsadam:
## Description of changes

Removed a full page to create a new user group and added a modal on the listing page.

## Related issue

N/A

## Checklist

*   [x] Tests added (or description of how to test) for any new features.
*   ~User documentation updated for UI or technical changes.~


src/main/webapp/resources/js/apis/users/groups.js
```diff
@@ -104,3 +104,13 @@ export function removeMemberFromUserGroup({ groupId, userId }) {
 export function getProjectsForUserGroup(groupId) {
   return axios.get(`${BASE_URL}/${groupId}/projects`).then(({ data }) => data);
 }
+
+export function createUserGroup({ name, description }) {
```
deepsidhu85, line 108: Comments

src/main/webapp/resources/js/components/PageLoadingIndicator.jsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
```
deepsidhu85, line 1: Maybe we can just reuse the `ContentLoading` component and update it as need be instead of having 2 similar components

deepsidhu85:
When creating a new group and a group with the name already exists the following error is thrown in the console:

Uncaught (in promise) TypeError: Cannot read property 'id' of undefined at CreateNewUserGroupButton.jsx:40

src/main/webapp/resources/js/apis/users/groups.js
joshsadam: Updated in d67ddf17ef3c5c40268ce4405320c918192bcf93

src/main/webapp/resources/js/components/PageLoadingIndicator.jsx
joshsadam: Makes sense, updated in 9225af5cf70fa5658f9c10b64307c6b58061ae7c

joshsadam:
Fixed last error in 8805a04

PR #708, opened 2020-05-26
Disbale fastqc zipped

tom114:
## Description of changes
Disabling the FastQC file processor for zipped fast5 files (`.fast5.tar.gz`).  Previously it was failing "gracefully", but that was creating qc entries for the failure even though it was working as intended.  Now it just skips the processor for that file type.

## Related issue
N/A

## Checklist

### DIFF
--- a/src/main/java/ca/corefacility/bioinformatics/irida/processing/FileProcessor.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/processing/FileProcessor.java
@@ -40,12 +40,12 @@ public interface FileProcessor {
 	/**
 	 * This method asks the file processor whether it should act on this file.
 	 * The processor may have some settings that would not run on certain files.
-	 * 
-	 * @param sequencingObjectId
-	 *            the {@link SequencingObject} id to check
+	 *
+	 * @param sequencingObject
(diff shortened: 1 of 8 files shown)


src/main/java/ca/corefacility/bioinformatics/irida/processing/FileProcessor.java
```diff
@@ -40,12 +40,12 @@
...
-	 * 
-	 * @param sequencingObjectId
-	 *            the {@link SequencingObject} id to check
+	 *
+	 * @param sequencingObject
+	 *            the {@link SequencingObject} to check
 	 * @return true if the processor should act on the file
 	 */
-	public default boolean shouldProcessFile(Long sequencingObjectId) {
+	public default boolean shouldProcessFile(SequencingObject sequencingObject) {
```
tom114, line 48: This wasn't used anywhere except tests yet. It works perfectly for this use case, but I changed it to accept the object instead of the id.  That made sense seeing it was being read in the `DefaultFileProcessingChain` right after anyway.

joshsadam: Looks good, thanks

PR #735, opened 2020-06-17
Admin layout

paulowenbrown:
## Description of changes
Set up overall layout of the admin panel. This includes: a side menu, top navigation bar, and pages for Statistics, Users, and Group Users. Statistics will be fleshed out in a later PR, whereas Users and Group Users had been designed previously and were lazy loaded in.

## Related issue
N/A

## Checklist


src/main/webapp/resources/js/pages/admin/components/Admin.jsx
```diff
@@ -0,0 +1,66 @@
...
+          <Content>
+            <Suspense fallback={<ContentLoading />}>
+              <Router>
+                <AdminStatistics
+                  path={
+                    `${DEFAULT_URL}/${ADMIN.STATISTICS}`
+                  }
+                  default
+                />
+                <AdminUsers
```
deepsidhu85, line 49: Since we are just lazy loading the components in and don't need to add anything to the page, you might be able to get rid of the `AdminUsers.jsx` and `AdminUserGroups.jsx` files and instead import `AdminUserGroupsPage` and `AdminUsersPage`.  Then here you would replace `AdminUsers` with `AdminUsersPage` and `AdminUserGroups` below with `AdminUserGroupsPage`.

src/main/webapp/resources/js/pages/admin/components/AdminSideMenu.jsx
```diff
@@ -0,0 +1,61 @@
...
+                    {i18n("admin.panel.statistics")}
+                  </Link>
+                </Menu.Item>
+                <SubMenu key="users" title={i18n("admin.panel.users")} className={"t-admin-users-submenu"}>
+                  <Menu.Item key="userList">
+                    <Link to={`${DEFAULT_URL}/${ADMIN.USERS}`} className={"t-admin-users-link"}>
+                      {i18n("admin.panel.userList")}
+                    </Link>
+                  </Menu.Item>
+                  <Menu.Item key="groupList">
```
deepsidhu85, line 48: Rename the key to userGroupsList

src/main/webapp/resources/js/pages/admin/routes.js
```diff
@@ -0,0 +1,10 @@
+/*
+ * This file contains the route names for the admin panel
+ */
+
+export const ADMIN = {
+  ADMIN: "admin",
+  STATISTICS: "statistics",
+  USERS: "users",
+  GROUPS: "groups"
```
deepsidhu85, line 9: Rename to UserGroups: "user-groups" and update the key wherever it is used

paulowenbrown: I am using these routes to change the URL. The old settings cog had /groups. Should it still be changed in this case? That means the path would then become admin/user-groups when the groups link is clicked

src/main/webapp/resources/js/pages/AdminUsersPage.jsx
```diff
@@ -147,7 +147,7 @@ function UsersTable() {
  * @returns {*}
```
joshsadam, line 147: Can you move this inside the admin directory, want to keep everything specifically tied to the admin in it's own grouping.

src/main/webapp/resources/js/pages/admin/components/AdminSideMenu.jsx
```diff
@@ -0,0 +1,61 @@
...
+  // The following renders the AdminPanelSideMenu component
+  return (
+    <Sider>
+      <Location>
+        {props => {
+          const keyname = props.location.pathname.match(pathRegx);
+          return (
+            <section>
+              <Link style={{ paddingLeft: 10, paddingRight: 10 }} to={`${DEFAULT_URL}/${ADMIN.STATISTICS}`}>
+                <img style={{ height: 64, width: 180 }} src="/resources/img/irida_logo_dark.svg"/>
```
joshsadam, line 33: The img src needs to be wrapped in `setBaseUrl`

src/main/resources/i18n/messages.properties
```diff
@@ -2150,4 +2150,12 @@ EditableParagraph.tooltip=Edit
 # ========================================================================================== #
 # ADMINISTRATIVE PANEL                                                                       #
 # ========================================================================================== #
-admin.panel.title=Admin Panel
\ No newline at end of file
+admin.panel.title=Admin Panel
+admin.panel.users=Users
```
joshsadam: Try to name your keys after the component they are tied to e.g. `AdminPanel.title` or whatever the component is.  That way it is easy to track down where it is used.  We have had problems deleting keys that get re-used.  I have also begun using a system that if it is internationalized on the server (e.g. the response from a request), preface the key with `server.`  just makes it easier to track down.

src/main/webapp/pages/admin/index.html
```diff
@@ -1,10 +1,28 @@
 <!DOCTYPE html>
-<html lang="en" xmlns:th="http://www.thymeleaf.org">
+<html lang="en"
+      xmlns:th="http://www.thymeleaf.org"
+      xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout">
```
joshsadam: Don't need \`xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout"\` it is used for doing complicated templating on the main site that I really wish was not there.

deepsidhu85:
This looks really good and works great!

src/main/webapp/entries.js
```diff
@@ -86,4 +86,5 @@ module.exports = {
   "user-details": "./resources/js/pages/user-details.js",
   "project-metadata-templates":
     "./resources/js/pages/projects/metadata-templates",
+  admin: "./resources/js/pages/admin/index.js",
```
joshsadam, line 89: FYI, and please don't bother changing it this is just a teaching point.  You don't have to include `index.js`, that is the default file name for js files so if you just had `admin: "./resources/js/pages/admin"` it would also be valid.

src/main/webapp/pages/admin/index.html
```diff
@@ -1,10 +1,29 @@
 <!DOCTYPE html>
-<html lang="en" xmlns:th="http://www.thymeleaf.org">
-<head>
+<html
+  lang="en"
+  xmlns:th="http://www.thymeleaf.org"
+>
+  <head>
     <meta charset="UTF-8">
```
joshsadam, line 7: Please add:

```html

	<meta http-equiv="X-UA-Compatible" content="IE=edge" />
	<meta name="viewport" content="width=device-width, initial-scale=1"/>
    <link rel="icon" href="/resources/img/favicon.ico" th:href="@{/resources/img/favicon.ico}" />
```

src/main/webapp/pages/admin/index.html
```diff
@@ -1,10 +1,29 @@
...
+  </head>
+  <body>
+    <div id="root" style="height: 100%;" layout:fragment="page">
+      <!--
+         This is a DOM element for React to render the admin details into.
+         See src/main/webapp/resources/js/pages/analysis/components/Admin.jsx
+      -->
+    </div>
+
+    <th:block layout:fragment="scripts">
```
joshsadam, line 19: You don't need the `<th:block>` wrapper here.

src/main/webapp/resources/js/pages/UsersPage.jsx
```diff
@@ -175,4 +176,7 @@ function UsersPage() {
   );
 }
 
-render(<UsersPage />, document.querySelector("#react-root"));
```
joshsadam, line 178: This is kind of ugly.  Can you move the \`UsersTable\` into it's own file, in the components directory (follow the naming in there for directory names).  Then create a `AdminUsersPage` that imports that.  The benefit is that it prevents checking the page variable if the user is an admin, you know if they are in the admin section that they must be one.  You can then pass that in as a param into the `UsersTable` component.

Issue #477, opened 2019-10-03
Remove angular ui


joshsadam:
## What needs changed?
Angular UI is used on multiple pages to handle thing like modals, dropdown menus, etc...  It is being removed from IRIDA as its major dependency is bootstrap.

## How does it work now?
Angular UI is a combination between angularjs and bootstrap allowing for dynamic creation of modals, dropdowns, etc...

## How should it work?
Replace with Ant Design.

## Additional information
This will need to be a part of a larger refactor.  Please add any pages that currently depend on this here.

tom114:
Partially done in #559

joshsadam:
This has been removed for a while now.  Could not find PR with it. No longer in `package.json`

PR #749, opened 2020-06-29
Remove older irida.cart.js angularjs service

joshsadam:
## Description of changes
Sorry about the formatting changes on pipeline.launch.js

This remove the very old angularjs cart service and replaces it with functions that are already existing within the front end cart api.

## Related issue
Related to #477 

## Checklist


src/main/webapp/resources/js/modules/cart/irida.cart.js
```diff
@@ -1,37 +0,0 @@
-import angular from "angular";
-import { CART } from "../../utilities/events-utilities";
-import { emptyCart, removeSample } from "../../apis/cart/cart";
-
-function CartService(scope, $http) {
```
tom114, line 5: I see the `removeSample` function reimplemented, what about the other service functions in here?  I'm assuming they're already done somewhere else?

src/main/webapp/resources/js/pages/pipelines/pipeline.launch.js
```diff
@@ -307,8 +307,12 @@ function PipelineController(
    * @param projectId the project id of the sample to remove
    * @param sampleId the sample if to remove
    */
-  vm.removeSample = function(projectId, sampleId) {
-    CartService.removeSample(projectId, sampleId).then(function() {
+  vm.removeSample = function (projectId, sampleId) {
```
tom114, line 310: So I'm assuming this is just changing to the cart functions in `resources/js/apis/cart/cart.js`?

joshsadam: Yes that is all that is happening here.

src/main/webapp/resources/js/modules/cart/irida.cart.js
joshsadam: So the answer to all your questions is that these functions had already been re-written for updates to other features (mostly the new cart).  The only functionality that was really still used was remove sample from cart I just refactored to use the new one.

PR #764, opened 2020-07-27
Hotfix sample metadata fields 

joshsadam:
## Description of changes

Fixes issues raised by @nknox:

1.  Numeric fields imported from an excel spreadsheet need to keep their formatting. **Fixed by checking during file upload if it is a numeric field.  If it is, it saves the value as it is formatted in the cell.**
2.  Long values do not wrap properly on the sample details page. **Fixed by adding a word-break css attribute to the `dd` tags.**

## Related issue

N/A

## Checklist

*   ~Tests added (or description of how to test) for any new features.~
*   ~User documentation updated for UI or technical changes.~


src/main/java/ca/corefacility/bioinformatics/irida/ria/web/projects/metadata/ProjectSampleMetadataController.java
```diff
@@ -132,14 +132,22 @@ public SampleMetadataStorage createProjectSampleMetadata(HttpSession session, @P
...
 					if (columnIndex < headers.size()) {
 						String header = headers.get(columnIndex);
 
 						if (!Strings.isNullOrEmpty(header)) {
 							// Need to ignore empty headers.
-							cell.setCellType(CellType.STRING);
-							rowMap.put(header, cell.getStringCellValue());
+							if(cell.getCellTypeEnum().equals(CellType.NUMERIC)) {
+								DataFormatter formatter = new DataFormatter();
+								String value = formatter.formatCellValue(cell);
```
tom114, line 144: This seems to work for me, glad it was that straightforwad.  Even rounding seems to happen the same as expected.  Can you add a comment about why it's being added?

src/test/java/ca/corefacility/bioinformatics/irida/ria/integration/pages/projects/ProjectSampleMetadataImportPage.java
```diff
@@ -55,4 +58,15 @@ public int getFoundCount() {
 	public int getMissingCount() {
 		return Integer.parseInt(missingPill.findElement(By.className("badge")).getText());
 	}
+
+	public List<String> getNumberColumnValues() {
```
tom114, line 62: This method is a bit too specific to the test data being used in `ProjectSampleMetadataImportPageIT.java`.  Can you change it to accept the header name as a string so it can be passed in by the test?  That way we can better reuse this page object if we wanted different input data.

joshsadam: Updated

PR #790, opened 2020-09-02
Fix OAuth2 redirect url builder

### LINKED ISSUE #677: Trailing '/' in server.base.url causes remote connections to fail
## Describe the bug
I found that if I have a trailing `/` in the value for `server.base.url` (from the `/etc/irida/web.conf` file) then remote connections will fail to connect.

For example, say I had `server.base.url=http://localhost:8080/`. Then, when I go to connect the remote instance of IRIDA I get the following exception:

```
04 May 2020 10:09:00,033 DEBUG ca.corefacility.bioinformatics.irida.web.controller.api.exception.CustomOAuth2ExceptionTranslat
or:51 - Passing exception error="invalid_grant", error_description="Redirect URI mismatch." to super to handle.
04 May 2020 10:09:00,070 ERROR ca.corefacility.bioinformatics.irida.ria.web.errors.ExceptionHandlerController:98 - OAuth excep
tion: invalid_grant, Redirect URI mismatch.
OAuthProblemException{error='invalid_grant', description='Redirect URI mismatch.', uri='null', state='null', scope='null', redirectUri='null', responseStatus=0, parameters={}}
        at org.apache.oltu.oauth2.common.exception.OAuthProblemException.error(OAuthProblemException.java:59)                         at org.apache.oltu.oauth2.client.validator.OAuthClientValidator.validateErrorResponse(OAuthClientValidator.java:63)
        at org.apache.oltu.oauth2.client.validator.OAuthClientValidator.validate(OAuthClientValidator.java:48)                        at org.apache.oltu.oauth2.client.response.OAuthClientResponse.validate(OAuthClientResponse.java:64)
        at org.apache.oltu.oauth2.client.response.OAuthClientResponse.init(OAuthClientResponse.java:59)                               at org.apache.oltu.oauth2.client.response.OAuthAccessTokenResponse.init(OAuthAccessTokenResponse.java:52)
...
```

However, if I change `server.base.url=http://localhost:8080` (no trailing `/`) it connects just fine.

1. Set `server.base.url=http://localhost:8080/` (trailing `/`)
2. Setup a client on this IRIDA instance (with redirect uri = `http://localhost:8080/api/oauth/authorization/token`). This should use the `authorization_code` grant type.
3. Setup a remote connection to this same IRIDA instance with `http://localhost:8080/api`.
4. Attempt to connect the remote connection. After clicking `Authorize` the connection will fail with the above exception.

## Expected behaviour
I expected to be able to connect whether or not the `server.base.url` has a trailing `/`

## Additional context
None.

tom114:
## Description of changes
Fixed an issue where paths were being appended by string concatenation in the OAuth2 flow instead of using a URI builder.  If someone had a trailing slash in the `server.base.url` (ex `http://localhost:8080/`) it could cause a double slash in the redirect URI.  This would cause the OAuth2 redirect verification to fail.

To test, in your `/etc/irida/web.conf` file try adding a slash to the end of the `server.base.url`, then using the test database try connecting to your own machine by a remote API.

## Related issue
Fixes #677 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Changes
 * [UI]: Fixed bug where import bulk metadata button was displayed on linelist page for project collaborators. (20.05.04)
 * [UI]: Fixed bug where add member to project button was displayed on project members page for project collaborators. (20.05.04)
 * [Developer]: Added optional method `getAnalysisViewer` for IridaPlugin to allow external applications to use IRIDA's built in analysis visualizations (such as the phylogenetic tree viewer).
+* [REST]: Fixed issue with double slashes sometimes getting added in OAuth2 redirect URI causing auth flow to fail.
 
 20.01 to 20.05
 --------------
--- a/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/oauth/OltuAuthorizationController.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/oauth/OltuAuthorizationController.java
@@ -25,8 +25,6 @@ import ca.corefacility.bioinformatics.irida.service.RemoteAPITokenService;
 
 /**
  * Controller for handling OAuth2 authorizations
- * 
- *
  */
 @Controller
 public class OltuAuthorizationController {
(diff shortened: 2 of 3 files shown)


deepsidhu85 (approved):
Works as expected. Thanks

PR #793, opened 2020-09-07
Update assembly pipeline (which finishes migrating all tools away from IRIDA toolshed)

### LINKED ISSUE #58: tbl2asn/Prokka self-destructs after a certain date
Hey,

So, Phil warned me about this a while ago, but I hadn't tried to address it.  We're using Prokka for annotations, which uses `tbl2asn` from NCBI.  `tbl2asn` self-destructs after a year and refuses to run.

Phil had been addressing this by manually re-downloading and swapping out the binary on Galaxy.  However, this messes up with reproducibility since we either don't keep track of the exact version of tbl2asn (swap it out on our own and not update the Galaxy tool), or we update our workflow for annotation every year and just leave old versions unable to complete.

I was reminded of this now because we've hit that point:

```
Fatal error: Exit code 2 ()
[tbl2asn] This copy of tbl2asn is more than a year old.  Please download the current version.
[16:53:38] Could not run command: tbl2asn -V b -a r10k -l paired-ends -M n -N 1 -y 'Annotated using prokka 1.10 from http://www.vicbioinformatics.com' -Z outdir\/prokka\.err -i outdir\/prokka\.fsa 2> /dev/null
```

Thoughts @franklin.bristow and @thomas.matthews .  Not sure the best way to address this, but from now on all assemblies/annotations in IRIDA will fail.

Imported from GitLab issue #315.  Originally posted on 2015/09/04 11:07AM

### LINKED ISSUE #92: Update all pipelines to use conda for dependency management
Imported from GitLab issue #478.  Originally posted on 2017/06/22 10:40AM

### LINKED ISSUE #231: Perl dependancies are missing.
I've used the irida auto tool installer to configure my galaxy with all the irida pipelines. 
I've found a missing perl dependencies. The installer says everything is installed fine; so it would be good if it either 

1. installed the perl libraries
2. warned me that I need to install these libraries.

So far it is:

- autodie and Parallel::ForkManager, & ReadOnly  for the snvlyphml pipeline
- & Bioperl for the assembly/annotation pipeline.

apetkau:
### No more dependency of IRIDA Galaxy toolshed

This updates the **Assembly and Annotation** and **Assembly and Annotation Collection** pipeline to only use tools from the main Galaxy toolshed. This completes the task of switching away from the IRIDA toolshed and to the main Galaxy toolshed.

### Fix tbl2asn expiry

This also updates some of the tools in the assembly pipelines. In particular, this updates `prokka` to a version where a dependency tool `tbl2asn` will no longer expire. This works by having prokka depend on the [tbl2asn-forever](https://github.com/audy/tbl2asn-forever) project in conda which uses `libfaketime` to fake the system time (hence tbl2asn never expires).

### Update Galaxy Docker

This also updates the IRIDA Galaxy Docker to Galaxy version `20.05` and includes all the updated tools.

### Testing

To test this out you can first run the following Docker command to start up Galaxy:

```
docker run --name galaxy-irida -d -p 48888:80 -v /path/to/irida/data:/path/to/irida/data phacnml/galaxy-irida-20.05
```

This should include all the most recent tools already installed, except for MentaLiST. To test MentaLiST you will have to install yourself in Galaxy once Galaxy Docker is started up (you may have to install multiple times as it still doesn't install properly even from the web UI). I added instructions for this in the `UPGRADING.md` guide.

Note that the updated Galaxy now uses the following account credentials:

* __Username:__ `admin@galaxy.org`
* __Password:__ `password`
* __API Key:__ `fakekey`

## Related issue

* Fixes #92 (all pipelines now use conda)
* Fixes #231  (these issues with Perl dependencies missing should be fixed)
* Fixes #58 (tbl2asn expiry issue)

## Checklist


tom114 (approved):
Tested with Galaxy 20.01, 20.05, and the referenced Docker.  All installed and completed successfully.  Thanks, merging.

PR #833, opened 2020-10-20
Launch parameters with options

joshsadam:
## Description of changes
Add pipeline parameters that have options.  Good examples are both the snvphyl and bio_hanself  pipelines.

 - If the parameter is boolean, it will be rendered with a checkbox
 - If the parameter has less than 7 options, it will be rendered as radio inputs
 - If the parameter has more than 7 options, it will be rendered as a select input

I added some JavaScript tests because there are a number of data transformations that occur in the UI and I wanted to make sure they were correct.  You can run them by running `yarn run test resources/js/pages/launch/launch-utilities.test.js` from the webapp directory.

## Related issue
N/A

## Checklist


deepsidhu85 (approved):
Nicely done!

PR #849, opened 2020-11-03
Add html result viewer

### LINKED ISSUE #164: Preview HTML analysis output files in iframe?
For example, QUAST Icarus contig size viewer:

image

Or QUAST HTML report:

image

Imported from GitLab issue #652.  Originally posted on 2018/05/30 10:26AM\Posted by Peter Kruczkiewicz

deepsidhu85:
## Description of changes

Add html result viewer for analyses that output html files

**As none of the pipelines within IRIDA output single html files we must do the following to test it**

1) Get the files from `irida-sample-data` available at https://irida.corefacility.ca/downloads/data/irida-sample-data.zip. Within there go into the `miseq-run-salmonella directory` and upload the 2 files (`AE014613_S1_L001_R1_001.fastq.gz` and `AE014613_S1_L001_R2_001.fastq.gz`) into a sample.
2) Run the `SISTR` analysis pipeline
3) Once it's complete, create an `html` file using a text editor with a basic heading etc in it
4) In the mysql database > `analysis_output_file table` update the `file_path` for one of the output files from the analysis above. For example if a file is saved as `26/0/FILENAME.EXT` then update it to `26/0/FILENAME.html`
5) Move your html file to the path in the above step. The path where the file should be is `/opt/irida/data/output/26/0/`.

Note the `26` in the path above will be different. Once the sistr analysis completes, within your terminal/intellij console you should see which paths the output files were copied to. 

## Related issue

Fixes #164 

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
```diff
@@ -151,4 +162,47 @@ public String getAdvancedPhylogeneticVisualizationPage(@PathVariable Long submis
...
+					.getName()
+					.contains(filename) && FilenameUtils.getExtension(filename)
+					.equals(htmlExt)) {
+				outputFile = file;
+				break;
+			}
+		}
+
+		try {
+			htmlOutput = FileUtils.readFileToString(outputFile.getFile()
```
tom114, line 197: An issue I can see with this is if the HTML file is really big.  I've seen some tools which may (for example) dump a pile of data in a javascript object for displaying a result on a page.  In a case like this loading everything into a string in memory could cause a problem.

Alternative: can we stream this output directly as a response?  I think you can do something like that but I can't find an example in our existing codebase.  I know you can get a handle on the request & response objects in the controllers and write to them similar to what we do in `FastqView`.  Can you check if that's possible for these purposes as well?

tom114 (changes requested):
Thanks, on my testing this seems to work!  Got a question below about how we can stream a file result instead of loading into memory.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
deepsidhu85: I can for sure take a look into if we can do that! Thanks :)

deepsidhu85: Updated in f9123a19284ed01987451861e3da27613ae05b86

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
```diff
@@ -151,4 +167,59 @@ public String getAdvancedPhylogeneticVisualizationPage(@PathVariable Long submis
...
+			}
+		}
+
+		// Set the common Http headers
+		response.setHeader(HttpHeaders.CONTENT_DISPOSITION, "inline");
+		response.setHeader(HttpHeaders.CONTENT_TYPE, "text/html");
+
+		try (InputStream inputStream = new FileInputStream(outputFile.getFile()
+				.toString()); OutputStream outputStream = response.getOutputStream()) {
+			response.setHeader(HttpHeaders.CONTENT_LENGTH, String.valueOf(Files.size(outputFile.getFile())));
```
tom114, line 206: I think the browser will handle things without this response header.  It looks like it should just stream the response.  That way if there's some filesystem size hiccup here it won't make the browser freak out that the content is different than the size it recieved.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
```diff
@@ -151,4 +167,59 @@ public String getAdvancedPhylogeneticVisualizationPage(@PathVariable Long submis
...
+					new Object[] { filename }, locale);
+			OutputStream outputStream = response.getOutputStream();
+			/*
+			Write the htmlNotFound message to the outputstream. We do this
+			so that the page doesn't error and will instead display the
+			message.
+			 */
+			outputStream.write(htmlOutputNotFound.getBytes(StandardCharsets.UTF_8));
+			outputStream.flush();
+			outputStream.close();
```
tom114, line 222: The output stream will get closed anyway with the try-with-resources block right?  Is this needed?

tom114 (changes requested):
Thanks for adding the stream handling.  A couple questions/changes below otherwise looking good!

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
deepsidhu85: Sounds good I will remove that response header

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
deepsidhu85: Since we reaccess it in the catch block would it still close? Or will it just close the original access stream?

tom114: Ah I'm sorry, I misread the code.  I think you have it correct.  Since you got a handle on a new stream a couple lines up you'll have to close it.  I thought for some reason that this was the same stream you opened in the try-resources block.  This is good and can stay as-is.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
deepsidhu85: Updated in 1c116df5695ce8f3b89c8013f76a705e6dd3482f

tom114 (approved):

apetkau:
What it looks like to integrate HTML output from Galaxy inside IRIDA:

image

PR #873, opened 2020-11-26
Add reference file upload progress

deepsidhu85:
## Description of changes

Reference file uploads now display progress for each file being uploaded to a project using a custom request.

## Related issue

## Checklist


src/main/webapp/resources/js/apis/projects/reference-files.js
```diff
@@ -41,3 +41,23 @@ export async function removeProjectReferenceFile(projectId, fileId) {
...
+  return await axios
+    .post(
+      setBaseUrl(`ajax/reference-files/project/${projectId}`),
+      formData,
+      config
+    )
+    .then(({ data }) => data)
+    .catch((error) => {
+      throw new Error(error.response.data.error);
+    });
```
joshsadam, line 62: When using `await` keep to the syntax you like (this is probably how I have been doing it and it is not exactly correct.  Keep the way I have seen you write it, it is much more elegant.  Also, I learned recently that instead of throwing a new error, just return a rejected promise:

```javascript
export async function uploadProjectReferenceFiles(projectId, formData, config) {
  try {
    const { data } = await axios.post(
      setBaseUrl(`ajax/reference-files/project/${projectId}`),
      formData,
      config
    );
    return data;
  } catch (e) {
    return Promise.reject(error.response.data.error);
  }
}
```
You can then handle this the same way:
```javascript
uploadProjectReferenceFiles(projectId, formData, config)
  .then(console.log)
  .catch(console.error);   
```

**You should start freaking out on me when I don't do it this way!**

deepsidhu85: Thanks. Yeah I've found myself flip flopping on it too. Oh for sure!

deepsidhu85: Updated in 496e77453248495380b0e3afe0eea9b5031b7f33

joshsadam (approved):
This looks fantastic, thanks

PR #901, opened 2020-12-31
Added hashing service to allow remote APIs to check for changes before launching full sync

### LINKED ISSUE #392: Add a hash on Project that covers all samples, files, etc within
Add a new field on a `Project` which only changes if anything inside the project changes (samples, files, metadata, etc).  This would likely have to be calculated on a scheduled basis as it would effectively have to do a similar scan to what the project sync job does, but instead it can do it all locally.  This would save a pile of time when remote services are checking for new data but a project has no changes as it can just check this code.  If their codes match, then no more requests needed.

REST API.  This would add a new data property to `Project`.

## Additional information
We had an issue with IRIDA when multiple remote installs were trying to sync data from multiple projects at the same time.  The current process is to check the `Project` for any data field updates, then check all `Samples` within for changes, then check for any new files within `Samples`.  This can get hard on the server if too many things are running at the same time.

tom114:
## Description of changes
Added a new service and REST API endpoints which allow REST API clients to get a full hashcode of a project.  This is being used so that remote APIs can check if there's any changes to the project with 1 request rather than doing a deep inspection of every sample and field within the project.

The new sync process works by using the new `ProjectHashingService`.  This service looks at all project, sample, sequencingobject, assembly, and metadata objects within a project and computes a hashcode for the full project.  This is a intensive task, but is similar to what would be done for a project sync job already.  While it will add a small amount of extra time for a sync job with new changes, it will dramatically increase the speed of sync jobs where there are no changes as it can complete the sync with 1 REST request.

To mitigate issues where the remote stored hashcode and data gets out of sync, this also adds a `Force Full Project Sync` button on the project remote page to run a full sync job if it's thought there may be unsync'd changes.

## Related issue
Fixes #392 

## Checklist


src/main/webapp/resources/js/pages/projects/remote/ProjectSynchronizationSettings.jsx
```diff
@@ -46,21 +46,39 @@ export function ProjectSynchronizationSettings() {
...
+                    onClick={() => updateSyncSettings({ forceSync: true })}
+                    disabled={
+                      (disableSyncNow ? true : false) ||
+                      (syncNowEnabledStates.includes(
+                        remoteProjectData.remoteStatus.syncStatus
+                      )
+                        ? false
+                        : true)
+                    }
+                    className="t-sync-now-btn"
```
deepsidhu85, line 77: Change this className to `t-sync-force-btn`

deepsidhu85 (changes requested):
Just looking at the UI stuff currently. Update the className below for the `Force Full Project Sync` button. Also, add a check in `ProjectRemoteSettingsIT` to ensure that the force sync button is displayed

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectHashingService.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.stereotype.Component;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Component used to compute a deep hashcode of a full project.  Looks at all project details, samples, metadata, and sequencing data within the project to generate a full hashcode.
```
deepsidhu85, line 23: Break this comment up over multiple lines

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectHashingService.java
```diff
@@ -0,0 +1,94 @@
...
+    @PreAuthorize("hasPermission(#project, 'canReadProject')")
+    public Integer getProjectHash(Project project) {
+
+        HashCodeBuilder builder = new HashCodeBuilder();
+
+        builder.append(project);
+
+        List<Join<Project, Sample>> samplesForProject = sampleService.getSamplesForProject(project);
+
+        for (Join<Project, Sample> j : samplesForProject) {
```
deepsidhu85, line 55: For this and below. Please rename the variables (j, e) to something like join, entry etc for clarity

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectSynchronizationService.java
```diff
@@ -185,80 +186,101 @@ public synchronized void findMarkedProjectsToSync() {
...
+			List<Join<Project, Sample>> localSamples = sampleService.getSamplesForProject(project);
 
-		//get a list of all remote URLs in the project
-		Set<String> remoteUrls = readSamplesForProject.stream()
-				.map(s -> s.getRemoteStatus()
-						.getURL())
-				.collect(Collectors.toSet());
+			// get all the samples by their url
+			Map<String, Sample> samplesByUrl = new HashMap<>();
+			localSamples.forEach(j -> {
```
deepsidhu85, line 232: Rename variable `j` for clarity

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectSynchronizationService.java
```diff
@@ -185,80 +186,101 @@ public synchronized void findMarkedProjectsToSync() {
...
+					logger.warn("Sample " + sample.getId() + " is not a remote sample.  It will not be synchronized.");
+				}
+			});
+
+			//read the remote samples from the remote API
+			List<Sample> readSamplesForProject = sampleRemoteService.getSamplesForProject(readProject);
+
+			//get a list of all remote URLs in the project
+			Set<String> remoteUrls = readSamplesForProject.stream()
+					.map(s -> s.getRemoteStatus()
```
deepsidhu85, line 250: Rename variable `s` here and below

src/test/java/ca/corefacility/bioinformatics/irida/service/ProjectHashingServiceIT.java
```diff
@@ -0,0 +1,116 @@
...
+        Collection<SampleSequencingObjectJoin> sequencingObjectsForSample = sequencingObjectService.getSequencingObjectsForSample(sample);
+        SequencingObject sequencingObject = sequencingObjectsForSample.iterator().next().getObject();
+
+        sampleService.removeSequencingObjectFromSample(sample, sequencingObject);
+
+        Integer newHash = hashingService.getProjectHash(project);
+
+        assertNotEquals(originalHash, newHash);
+    }
+
```
deepsidhu85, line 115: Add a test for metadata changes for a sample as well if possible. If that part isn't good to go just leave a todo

deepsidhu85 (changes requested):
Looking good. Just a few changes below

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectHashingService.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param project the {@link Project} to generate a hashsum for
+     * @return the computed hashsum
+     */
+    @PreAuthorize("hasPermission(#project, 'canReadProject')")
+    public Integer getProjectHash(Project project) {
+
+        HashCodeBuilder builder = new HashCodeBuilder();
+
+        builder.append(project);
```
deepsidhu85, line 51: A question about the builder.append. Should we maybe have these calls in a try/catch block? In case there any exceptions are thrown by the hashcodebuilder? That way at least we can log the exception rather than it being swallowed up.

deepsidhu85 (changes requested):

deepsidhu85:
Another thing that I noticed when testing is that on the Project Synchronization Settings page if we click the `Check for Changes` button it does not disable that button and the `Force Full Project Sync` button unless we refresh the page. Also, once it's done checking for changes if the `Check For Changes` button is clicked, it should enable the 2 buttons again

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectHashingService.java
tom114: Done in 6e6bea3.  For some reason my ide kicked out my formatting settings.

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectHashingService.java
tom114: Done in 6e6bea3.

src/main/webapp/resources/js/pages/projects/remote/ProjectSynchronizationSettings.jsx
tom114: Done in ee5e5c5

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectSynchronizationService.java
tom114: Done in 87cf637

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectSynchronizationService.java
tom114: Done in 87cf637

src/main/java/ca/corefacility/bioinformatics/irida/service/remote/ProjectHashingService.java
tom114: RIght now if an exception gets thrown while hashing it won't get swallowed, it'll go un-caught up to the top.  I think this is probably for the best since then it'll run up to the webserver which will give a 500 error and log the whole stack trace.  If there's any exceptions building the hash I think the right thing to do is to error out so I wouldn't want to just catch and log.

Other option would be to throw a new custom exception for this, but since there isn't really anything special to do to handle it I'm not sure it's worth it.  Thoughts?

deepsidhu85: Oh ok well as long as the error is swallowed then it's all good! Yeah not worth it in that case. Thanks

src/test/java/ca/corefacility/bioinformatics/irida/service/ProjectHashingServiceIT.java
tom114: added in 6529954

tom114:
Thanks for mentioning the disabling of the buttons.  I must have missed the other as I was testing.  Added disabling the button on either click in c00cc84

deepsidhu85:
> @deepsidhu85 Thanks for mentioning the disabling of the buttons. I must have missed the other as I was testing. Added disabling the button on either click in [c00cc84](https://github.com/phac-nml/irida/commit/c00cc84797bd2cc62041d98ca843704c8c64a34f)

All good! Thanks for updating it!

deepsidhu85 (approved):

PR #915, opened 2021-01-14
Fix issue with clearing samples from the cart

joshsadam:
## Description of changes

Currently if you:

1.  Add samples to the cart
2.  Go to the cart page and click the "Empty Cart" button on the bottom of the right side panel, the samples get removed
3.  If you try to add those same samples, you get an error that the samples cannot be added because a sample of the same name already exists in the cart.

This happens because the "Cart" extends HashMap and I only cleared the map, what I forgot was that there was a HashMap inside that contains the sample id and sample names for fast look ups (for seeing if the name is already in the cart).  This PR clears that when the cart is empty.  As well I added a bunch of testing for the cart service itself.

## Related issue

N/A

## Checklist

*   [ ] ~CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change.~ CHANGELOG does not need to be updated because this is only in development.
*   [x] Tests added (or description of how to test) for any new features.
*   [ ] ~User documentation updated for UI or technical changes.~

### DIFF
--- a/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/services/UICartService.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/services/UICartService.java
@@ -125,7 +125,7 @@ public class UICartService {
 	 * Remove all samples from the cart
 	 */
 	public void emptyCart() {
-		cart.clear();
+		cart.empty();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


ksierks:
Bug fixed! Looks good to me!

ksierks (approved):

PR #928, opened 2021-01-29
Changed git ignore for apidocs down to the main gitignore file

tom114:
## Description of changes
Moved ignoring the apidocs from the `/docs` directory gitignore down to the project root gitignore.  On deploying the new docs it was accidentally ignoring the apidocs changes.

## Related issue
N/A

## Checklist

### DIFF
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -1,4 +1,3 @@
 _site
 .sass-cache
 vendor
-developer/apidocs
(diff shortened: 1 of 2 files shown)


doc/.gitignore
```diff
@@ -1,4 +1,3 @@
 _site
 .sass-cache
```
joshsadam, line 2: Is the plan to get ride of this secondary .gitignore file and just keep that one in root?

tom114: No, this whole directory ends up getting copied over to the https://github.com/phac-nml/irida-documentation repo when we do a release.  The issue was that since the apidocs directory was within the gitignore here, it wasn't updating the apidocs when we released the new doc code.  We want to keep some things ignored (like the `_site` stuff).

joshsadam: Sounds good. This looks great then thanks

joshsadam (approved):

Issue #941, opened 2021-02-13
“TypeError: can't convert undefined to object” in Advanced Visualisation - 


zipho:
## Describe the bug

When I added metadata/linelist on my samples.

The advanced visualisation page hangs on load with an error from the browser console.

“TypeError: can't convert undefined to object” 

It was only later when I realise that it only occurs when the reference used to construct the newick tree has anything other than `>reference` at the first line.

For my mtb pipeline mine had `>Chromosome` and sometimes `>NC_000962.3`. 

## Steps to reproduce the problem
What were you doing when you encountered the problem?
1. Run a phylogeny pipeline on IRIDA
2. Upon results Viewed Results 
3. Clicked on Advanced Visualisation

## Expected behaviour
For my phylogeny tree to be visible with metadata.
![Screenshot from 2021-02-13 08-42-46](https://user-images.githubusercontent.com/2108971/107843891-96c7ba00-6dd7-11eb-968a-96c23a839fcb.png)

## Additional context
Latest IRIDA

tom114:
Fixed in #941

PR #942, opened 2021-02-13
Metadata Phylogeny-tree Advanced Visualisation fix 

zipho:
## Description of changes
If you use any reference file starting with line, other than `>reference`,  the for loop which adds data to the leaves of the tree would throw an error `can't convert undefined to object.`, basically happens when it tries to set `leaf.data = metadata['anything other than reference'];` - the result would be that `leaf.data = undefined` and the phylocanvas plugin would break when trying to run `tree.load(newick)`.

I  added an `if statement` to check the existence of leaf.label as a key in metadata object, which is the case for the samples and when it doesn't, as the case of a custom named reference e.g `>Chromosome`/`NC_000962.3` for my case, it would lookup the contents of `metadata.reference`.

## Related issue
https://github.com/phac-nml/irida/issues/941

Screenshot from 2021-02-13 09-12-07

### DIFF
--- a/src/main/webapp/resources/js/pages/visualizations/phylogenetics/components/phylocanvas/phylocanvas.component.js
+++ b/src/main/webapp/resources/js/pages/visualizations/phylogenetics/components/phylocanvas/phylocanvas.component.js
@@ -122,13 +122,13 @@ function phylocanvasController(
     // Add the metadata to the branches before the tree is drawn.
     tree.on("beforeFirstDraw", () => {
       for (const leaf of tree.leaves) {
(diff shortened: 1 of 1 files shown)


joshsadam:
Can you add the fix to the CHANGELOG.md file please.

zipho:
> Can you add the fix to the CHANGELOG.md file please.

I added `* [Developer]: Fixed the advanced visualisation viewer for Phylogeny Tree with metadata.` line.

joshsadam (approved):
Thanks @zipho this looks good.

PR #948, opened 2021-02-19
Added some args to run.sh that I use often

JeffreyThiessen:
## Description of changes
run.sh now defaults to the following
* dev profile
* skip yarn build
* do not rebuild db

adds the following commands to do other things
* `--prod` will run in prod profile
* `--yarn` will force it to build yarn (needed for first time running)
* `--create-db` will force it to remake database (needed for first time running)

This could use feedback on other args people use frequently to build this out a bit more to save time. The way I have set it up is that without any args it works as a quick launch into dev mode.

## Related issue
N/A

## Checklist

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,13 +1,42 @@
 #!/bin/bash
 export MIN_JS=false
 
-ADD=""
+PROFILE="dev"
+CREATE_DB=""
+YARN="-Dskip.yarn"
 
-if [ "$1" = "--create-db" ]; then
-   ADD="-Dhbm.dev.auto=create"
+for arg in "$@"
+do
+    if [ "$arg" = "--create-db" ]; then
+       CREATE_DB="-Dhbm.dev.auto=create"
+    fi
+
+    if [ "$arg" = "--yarn" ]; then
+       YARN=""
+    fi
+
+    if [ "$arg" = "--prod" ]; then
+       PROFILE="prod"
+    fi
+done
+
+
+if [ "$CREATE_DB" = "-Dhbm.dev.auto=create" ]; then
    echo "Dropping then Creating/Recreating database schema"
 else
    echo "Updating database schema without dropping"
   fi
 
-mvn clean jetty:run -Dspring.profiles.active=dev ${ADD}
+if [ "$YARN" = "" ]; then
+   echo "Building Yarn"
+else
+   echo "Continuing without yarn"
+  fi
+
+if [ "$PROFILE" = "prod" ]; then
+   echo "Running with Production Profile"
+else
+   echo "Running with Development Profile"
+  fi
+
+mvn clean jetty:run -Dspring.profiles.active=${PROFILE} ${CREATE_DB} ${YARN}


run.sh
```diff
@@ -1,13 +1,42 @@
 #!/bin/bash
 export MIN_JS=false
 
-ADD=""
+PROFILE="dev"
+CREATE_DB=""
+YARN="-Dskip.yarn"
```
tom114, line 6: @joshsadam  might have opinions here, but I feel like the default should be to *not* skip yarn.  I feel like it's pretty regular that a UI change could have been made, so it should run by default.  Plus on the first run after cloning the repo I think that yarn *must* run so we don't want to make it a non-default.

run.sh
```diff
@@ -1,13 +1,42 @@
...
+   echo "Continuing without yarn"
+  fi
+
+if [ "$PROFILE" = "prod" ]; then
+   echo "Running with Production Profile"
+else
+   echo "Running with Development Profile"
+  fi
+
+mvn clean jetty:run -Dspring.profiles.active=${PROFILE} ${CREATE_DB} ${YARN}
```
tom114, line 42: Any way to add any non-recognized arguments to the end of this line?  That way if someone adds some additional mvn arguments they can get added here.

run.sh
```diff
@@ -1,13 +1,42 @@
...
+       CREATE_DB="-Dhbm.dev.auto=create"
+    fi
+
+    if [ "$arg" = "--yarn" ]; then
+       YARN=""
+    fi
+
+    if [ "$arg" = "--prod" ]; then
+       PROFILE="prod"
+    fi
```
tom114, line 20: If you're running in the `prod` profile you should also be setting `irida.db.profile=prod`

run.sh
```diff
@@ -1,13 +1,42 @@
 #!/bin/bash
```
tom114, line 1: If this is getting updated can you add a small usage comment at the top of the script to say what the args are for?

tom114: Also would need to be updated in the docs at https://phac-nml.github.io/irida-documentation/developer/getting-started/#running-and-building-irida

tom114 (changes requested):
I like some of the changes, thanks.  Anything we can make easier for devs and others is good

run.sh
joshsadam: These seem backwards to be.  If I added `--yarn` to the command line I would expect yarn to be run (or is it me that is backwards).  As to answer @tom114 above I think by default it is fine to ignore because it is so much faster.  Just make sure that it is noted somewhere that you need to run `yarn install` and one of `yarn build` or `yarn start`

run.sh
joshsadam: I think by default it is fine to ignore because it is so much faster.  Just make sure that it is noted somewhere that you need to run `yarn install` and one of `yarn build` or `yarn start`

ericenns:
Closing due to inactivity and solution implemented in #1188

PR #951, opened 2021-02-25
Fixed missing filter menus on analysis outputs table

### LINKED ISSUE #899: Missing button for filtering results in analysis output downloads
## Describe the bug

In IRIDA version `19.05.1` there existed a button you could use to filter the analysis output files displayed in a table:

image

However, in IRIDA `20.09.4` this button no longer exists:

image

1. In an older version of IRIDA (`19.05.1`) go to a particular project.
2. Within the project go to **Analyses > Shared Single Sample Analysis Outputs**
3. Hover over the table column headers (**File** or **Analysis Type**) and you should be given the option to filter the table.
4. Repeat steps 1-3 in a newer IRIDA (`20.09.4`) and the option to filter the table is missing.

## Expected behaviour
I expected to be able to filter the table of analysis results based on user-entered strings.

## Additional context
None.

joshsadam:
## Description of changes

Added the missing filter menus to the analysis outputs table. Added a specific date filter to the date column

## Related issue

Fixes #899 

## Checklist

*   [ ] ~Tests added (or description of how to test) for any new features.~
*   [ ] ~User documentation updated for UI or technical changes.~

### DIFF
--- a/src/main/webapp/resources/js/pages/analyses/analysis-outputs-table.js
+++ b/src/main/webapp/resources/js/pages/analyses/analysis-outputs-table.js
@@ -340,10 +340,25 @@ async function getTableData(isShared = true) {
       filterParams: {
         // provide comparator function
         comparator: (filterLocalDateAtMidnight, cellValue) => {
-          const cellDate = new Date(cellValue);
-          if (cellDate < filterLocalDateAtMidnight) {
+          const cellDateTime = new Date(cellValue);
+          const cellDate = new Date(
+            Date.UTC(
+              cellDateTime.getFullYear(),
+              cellDateTime.getMonth(),
+              cellDateTime.getDate()
+            )
+          );
+          const filterDate = new Date(
+            Date.UTC(
+              filterLocalDateAtMidnight.getFullYear(),
+              filterLocalDateAtMidnight.getMonth(),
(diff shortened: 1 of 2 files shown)


src/main/webapp/resources/js/pages/analyses/analysis-outputs-table.js
```diff
@@ -314,27 +317,41 @@ async function getTableData(isShared = true) {
...
+          cellRenderer: (p) => `${p.data.userFirstName} ${p.data.userLastName}`,
         }
       : null,
     {
       field: "createdDate",
       headerName: i18n("analysis.date-created"),
-      cellRenderer: p => formatDate({ date: p.data.createdDate })
-    }
-  ].filter(header => header !== null);
+      filter: "agDateColumnFilter",
```
deepsidhu85, line 339: The date filter is not working correctly.

`Equals` Not working at all
`Greater Than` Works but will also work for the `than` value. Say for example Greater than February 11th...It will also return the results for February 11th
`Not Equal` works but also returns the results for the date you want to exclude

deepsidhu85 (changes requested):
The analysis type filter is working great. However the date filter has some problems. See comment below

deepsidhu85 (approved):

PR #960, opened 2021-03-11
Fix for samples not rendering on project samples page when config property ngsarchive.linker.available set to false

### LINKED ISSUE #958: Setting the property ngsarchive.linker.available=false causes an error on the project_samples page
## Describe the bug
Setting the property `ngsarchive.linker.available=false` in the file `configuration.properties` causes the page that lists the samples of a project to go in error.  Apparently the file Linker.jsx needs the line `<span id="linker"></span>` in `project_samples.html`.

1. Set `ngsarchive.linker.available=false` in the file `configuration.properties`
2. Restart IRIDA
3. Open a project on the page listing its samples, the samples are not listed

## Expected behaviour
The page shows the samples of the project

ksierks:
## Description of changes
Added a conditional check that the html element exists before rendering the Linker component.

## Related issue
Fixes #958

## Checklist

* [ ] ~Tests added (or description of how to test) for any new features.~
* [ ] ~User documentation updated for UI or technical changes.~

### DIFF
--- a/src/main/webapp/resources/js/pages/projects/samples/linker/Linker.jsx
+++ b/src/main/webapp/resources/js/pages/projects/samples/linker/Linker.jsx
@@ -35,7 +35,7 @@ function Linker() {
 
   const options = [
     { label: i18n("Linker.fastq"), value: "fastq" },
-    { label: i18n("Linker.assembly"), value: "assembly" }
+    { label: i18n("Linker.assembly"), value: "assembly" },
   ];
 
   function handleSampleIds({ detail }) {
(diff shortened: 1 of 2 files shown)


src/main/webapp/resources/js/pages/projects/samples/linker/Linker.jsx
```diff
@@ -110,4 +110,7 @@ function Linker() {
   );
 }
 
-render(<Linker />, document.querySelector("#linker"));
+let linker_html = document.querySelector("#linker");
```
joshsadam, line 113: Don;t know about using snake case but I will give it to you here :)

joshsadam (approved):
This looks good

PR #974, opened 2021-04-07
Single request all project metadata

tom114:
## Description of changes
Added a REST endpoint to request a full project's sample metadata at the same time.  This should help with performance issues when reading large amounts of metadata at the same time.  This endpoint may be slow for large projects, but if you need to read a full project anyway should be faster than requesting for each individual sample.

## Related issue
N/A

## Checklist


tom114:
Closing as this is switching to a non-hotifx pr in #977

PR #990, opened 2021-05-06
Project settings  project redux api

joshsadam:
## Description of changes

Updated project details API to use Redux API

## Related issue

N/A

## Checklist

N/A


src/main/webapp/resources/js/apis/projects/project.js
```diff
@@ -0,0 +1,90 @@
...
+        },
+      }),
+      invalidatesTags: ["Project"],
+    }),
+    /*
+    Update the current default metadata template for a project
+     */
+    updateDefaultMetadataTemplate: build.mutation({
+      query: ({ projectId, templateId }) => ({
+        url: "/set-project-default",
```
deepsidhu85, line 72: Since this is now in the `ProjectDetailsAjaxController` we should update the url to `/set-project-default-template`

src/main/webapp/resources/js/apis/projects/projects.js
```diff
@@ -68,20 +40,3 @@ export async function updateProjectAttribute({ projectId, field, value }) {
 export async function getProjectInfo(projectId) {
```
deepsidhu85, line 40: Isn't this getting the same data as this code from `project.js`:

  `getProjectDetails: build.query({
    query: (projectId) => ({
      url: "",
      params: { projectId },
    }),
    providesTags: ["Project"],
  }),`
     
The only place using the `getProjectInfo` here is in ReferenceFile.jsx. I have not looked over the PR for that yet but if you have already fixed this there then it's alright

src/main/webapp/resources/js/pages/projects/settings/components/processing/ProcessingPriorities.jsx
```diff
@@ -12,8 +13,8 @@ import { putPriorityUpdate } from "../../../redux/projectSlice";
  * @constructor
  */
 export function ProcessingPriorities({ projectId }) {
-  const dispatch = useDispatch();
-  const { priority } = useSelector((state) => state.project);
+  const [updateProjectPriority] = useUpdateProjectPriorityMutation();
+  const { data: project } = useGetProjectDetailsQuery(projectId);
```
deepsidhu85, line 17: Shouldn't this be 

`const { data: project = {} } = useGetProjectDetailsQuery(projectId);`

deepsidhu85 (changes requested):
Looking great and works as expected. Just a few changes below

src/main/webapp/resources/js/apis/projects/project.js
joshsadam: Updated in 398c4bb

src/main/webapp/resources/js/apis/projects/projects.js
joshsadam: Yeah, just wait for that PR :)  You have a very good eye for details

src/main/webapp/resources/js/pages/projects/settings/components/processing/ProcessingPriorities.jsx
joshsadam: It doesn;t need to be but it also doesn't hurt.  I updated it in 4d5694d.  By the time this gets called here, it is already called in one of the other components, so this is just getting the cached data therefore does not have the opportunity to call for an attribute on undefiend.

deepsidhu85: Make sense. Thanks for updating it though :)

src/main/webapp/resources/js/apis/projects/projects.js
deepsidhu85: Sounds good. Thanks :)

deepsidhu85 (changes requested):
Everything else is good except a few changes below

src/main/webapp/resources/js/pages/projects/settings/components/metadata/MetadataTemplates.jsx
```diff
@@ -54,14 +53,13 @@ export default function MetadataTemplates({ projectId }) {
   const { data: fields } = useGetMetadataFieldsForProjectQuery(projectId);
   const [templates, setTemplates] = React.useState([]);
 
-  const { canManage, defaultMetadataTemplateId } = useSelector(
-    (state) => state.project
-  );
+  const { data: project } = useGetProjectDetailsQuery(projectId);
```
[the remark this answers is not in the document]
joshsadam: Updated in 67b61bd

src/main/webapp/resources/js/components/roles/ProjectRole.jsx
```diff
@@ -13,8 +13,8 @@ import { useRoles } from "../../contexts/roles-context";
  * @returns {*}
  * @constructor
  */
-export function ProjectRole({ item, updateRoleFn }) {
-  const { id: projectId, canManage } = useSelector((state) => state.project);
+export function ProjectRole({ projectId, item, updateRoleFn }) {
+  const { data: project } = useGetProjectDetailsQuery(projectId);
```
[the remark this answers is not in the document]
joshsadam: Updated in 67b61bd

deepsidhu85 (approved):

PR #993, opened 2021-05-06
Single query metadata

tom114:
## Description of changes
Created a custom query for retrieving a full project's metadata for the line list page.  The existing methods using hibernate/JPA were performing a ton of DB queries for retrieving metadata when a project had a high (1000+) number of samples.  This change grabs the metadata in a single(ish) query.

Performance testing reduces rendering time from ~10s for projects with 5k samples to ~1/4s.

## Related issue
Related to #991 

## Checklist
Things for the developer to confirm they've done before the PR should be accepted:

* [ ] CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change.
* [ ] Tests added (or description of how to test) for any new features.
* [ ] User documentation updated for UI or technical changes.

PR #997, opened 2021-05-10
Improving line list performance with custom metadata queries

### LINKED ISSUE #991: Improve line list performance
## Describe the bug
IRIDA's line list performance gets quite slow when a project contains a high number of samples and/or a high number of metadata columns.  Issue becomes noticeable when a project has 1000+ samples with metadata.

1. Add 1000+ samples with metadata to a project
2. Go to line list page
3. Wait for 10+ seconds for metadata to render

## Expected behaviour
Better performance for line list page.

## Additional context
Some performance improvements were realized in the 21.01 release which changed the way metadata was stored within the database, but issues still remain.  The solution for this will probably require custom queries to retrieve sample metadata.

tom114:
## Description of changes
Added a custom query to retrieve metadata from the database for displaying in the line list and REST metadata controller.  This has significantly reduced the query time for retrieving metadata on large projects (large both by number of samples and amount of metadata).  In some cases this is producing page load performance reductions from ~60s to ~1s.

## Related issue
Fixes #991 
More discussion in #993 and #994

## Checklist


ksierks (approved):
I ran IRIDA locally and imported the data from /Drives/T/tom/lotsofmetadata.sql. It took about 10 seconds for the metadata in Project 1 to load. After the change, it loads just as fast as any other page. That's pretty awesome!

ksierks (approved):

PR #999, opened 2021-05-13
Remote project default template fix

deepsidhu85:
## Description of changes

Fixed issue when creating a synchronized project with the project to be sync'd having a default metadata template set. Now the remote project has it's default_metadata_template set to null rather than having the value from the existing project.

**To test broken synchronized project functionality:**

1) Checkout development (from origin)
2) Create a project and add a sample
3) Run a pipeline
4) Create metadata template for project
5) Set this metadata template as the default for the project
6) Try to create a synchronized project.

Synchronized project should not be created and there should be an error in the terminal or if using intellij then there should be an error in the RUN console. The error should be something like `MetadataTemplate detached entity passed to persist`

**To test fixed synchronized project functionality:**

As a project with a sample was created as well as a metadata template was created and set as the default all you need to do is the following:

1) Checkout `remote-project-default-template-fix` from my fork (deepsidhu85)
2) Try to create a synchronized project.

The synchronized project should be successfully created. If you look at the database in the projects table you will see the synchronized project also has the default metadata template set to `NULL`.

## Related issue

## Checklist


deepsidhu85:
> I think I have a much easier fix to this issue, the `defaultMetadataTemplate` property doesn't even need to be visible through the rest api. Just add a `@JsonIgnore` to the property and I think all these problems go away.
> 
> Also I noticed that the import for the `@JsonIgnore` property is wrong in `Project`. Switch it to `import com.fasterxml.jackson.annotation.JsonIgnore;`

Awesome thanks! Updated in 8dc9aa244c84d79ab624b003b7e96e9581be507b

tom114 (approved):

Issue #931, opened 2021-02-03
Limit metadata fields in the line list by user's role on the project


tom114:
## Describe your idea for a new feature
Allow additional user roles to be set on a project, and those different roles may have different views of the metadata on the project.  Certain roles may have full access, some may have certain fields hidden to them, etc.

## Describe how you would access this feature?
In the line list and sample details views.

## Additional information
This will involve a fairly extensive modification to the line list, and particularly how we create and manage columns of data.  We will need to add something to restrict the columns by role in the UI, and also add it to the security layer to filter based on those roles.

The tasks for this issue are captured in more detail in https://github.com/phac-nml/irida/projects/12

tom114:
## Implementation plans:
Currently when managing sample metadata, all data is visible to all authorized viewers of the project.
![image](https://user-images.githubusercontent.com/8951708/113199689-07fae980-922d-11eb-97c4-ec3c5904852f.png)

The proposed metadata restrictions system will allow project managers to provide restrictions on a project so that portions of the metadata can be restricted to users with a given role on the project.

To build these functions, we'll use the existing `ProjectRole` system, but will define some new levels to add additional granularity.  A user's role on the project defines their permissions to data within the project.  Currently there exists 2 project roles in IRIDA: "Collaborator" and "Manager".  Manager has the ability to modify data on the project, where Collaborator only has view access to the project.  One of the initial tasks will be to add 1 or more additional roles to the project.  For these examples, we'll add the role "Level2" between the existing roles.

```
Roles:
* Level1 = "Collaborator"
* Level2 *new*
* Level3 = "Manager"
```

To set restrictions on a project, we'll add a new interface where all the available metadata fields on the project are listed.  This listing will include a "Restriction Level" column where project managers can set the visibility level of the fields within the project.  Users on the project must have >= the level defined in the restriction list in order to view the data in the column.

![image](https://user-images.githubusercontent.com/8951708/113200311-c3bc1900-922d-11eb-8854-fec6f5bd43af.png)

For a usage example, lets say we have 2 users with the given roles:
```
User1 = Level3
User2 = Level2
User3 = Level1
```

These restrictions would lead to the following visibility of columns depending on the user's project role:

![image](https://user-images.githubusercontent.com/8951708/113200448-f108c700-922d-11eb-84f3-cbb6b36e8613.png)

This will create a tiered presentation structure that changes the view of the project metadata based on the user's project role.
[...]

PR #1030, opened 2021-06-07
Restrict metadata responses

tom114:
## Description of changes
* Adds a function that asks which metadata fields a user is allowed to read.
* Adds to the function getting metadata entries.  Adds the ability to define which fields to get entries for.
* Added these 2 new functions to the line list controller to only show the allowed fields to the user.

Note: **These functions are not yet managed by spring security, they need proper permissions added**

## Related issue
Related to #931

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/service/impl/sample/MetadataTemplateServiceImpl.java
```diff
@@ -221,4 +235,79 @@ public MetadataRestriction setMetadataRestriction(Project project, MetadataTempl
...
+		//for each field to check
+		List<MetadataTemplateField> filteredFields = metadataFieldsForProject.stream()
+				.filter(field -> {
+					//if the restriction map contains the field
+					if (restrictionMap.containsKey(field)) {
+						MetadataRestriction metadataRestriction = restrictionMap.get(field);
+						ProjectRole level = metadataRestriction.getLevel();
+
+						//compare the restriction level to the given role.  If it's greater or equal, we're good
+						if (role.compareTo(level) >= 0) {
```
deepsidhu85, line 264: Just a question on this part. Does it just lexicographically compare the strings in the enum. Example say the level is "PROJECT_OWNER" and the role is "PROJECT_USER", does it just compare these strings to each other? If that's the case it works for now but we will need to update this when more project roles are added

deepsidhu85 (commented):
Just have a comment/question below. Looking great otherwise

src/main/java/ca/corefacility/bioinformatics/irida/service/impl/sample/MetadataTemplateServiceImpl.java
tom114: Java's compareTo for an enum will order the objects in the order they're declared in the enum <https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Enum.html#compareTo(E)>.  Currently in ProjectRole it's specified as:
```java
public enum ProjectRole {

    PROJECT_USER("PROJECT_USER"),
    PROJECT_OWNER("PROJECT_OWNER");
```
I'm pretty sure you can't override compareTo in an enum in Java so regardless of the string names it'll stay in order as long as we order them correctly in the enum code.

joshsadam: I think it might be more clear if we added a numeric value to the enum.

deepsidhu85: @tom114 Thanks for the clarification. I didn't know exactly what was being compared there. I agree with @joshsadam that adding a numeric value to the enum would add some clarity

tom114: Sure. Added a `level` property in ProjectRole so we can better compare those roles.  afc87f7

deepsidhu85: That's perfect. Thanks @tom114

joshsadam (approved):
This looks really great.  Things are definitely headed in the correct direction.  Thanks @deepsidhu85 for finding those few issues.

PR #1036, opened 2021-06-14
Refactor base permission

tom114:
## Description of changes
Refactored the `BasePermission` class to an interface rather than its original abstract class.  This way we can build permissions that aren't directly tied to a JPA repository in IRIDA.  Most of the internal parts of the old `BasePermission` are refactored down to `RepositoryBackedPermission` and all the old permissions now extend that class.

Functionally this branch should work exactly the same, this is just preparing for some more complex permissions for later.

## Related issue
Related to #931 

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/security/permissions/BasePermission.java
```diff
@@ -1,177 +1,29 @@
...
-	 */
-	private boolean customPermissionAllowedCollection(Authentication authentication, Collection<?> targetDomainObjects) {
-		boolean permitted = true;
-		for (Object domainObjectInCollection : targetDomainObjects) {
-			permitted &= customPermissionAllowedSingleObject(authentication, domainObjectInCollection);
-		}
-
-		return permitted;
-	}
+	String getPermissionProvided();
```
deepsidhu85, line 18: Shouldn't this have a access modifier infront of it such as public? Same thing below for `boolean isAllowed(Authentication authentication, Object targetDomainObject);`

deepsidhu85 (changes requested):
Everything appears to be working as it should. Just a comment below

src/main/java/ca/corefacility/bioinformatics/irida/security/permissions/BasePermission.java
tom114: Since BasePermission is now an interface, all methods inside will be `public`.  We have lots of interfaces in IRIDA that specify public, but they don't need to.

deepsidhu85: Sounds good. Thanks!

deepsidhu85 (approved):

PR #1095, opened 2021-09-20
Fixed metadata query error when project had no metadata

tom114:
## Description of changes
With the last query update the query failed when loading a line list for a project with no metadata:
```
MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')'
```

This was because if there was no metadata fields in the first new query, the `IN` clause in the second would be empty.  This PR adds a check to ensure that query is skipped if it'll end up returning no results.

## Related issue
N/A

## Checklist


ksierks (approved):
Sorry, I missed testing no metadata earlier. Works as expected.

PR #1144, opened 2021-12-09
Address single dynamic parameter in pipeline launch

joshsadam:
## Description of changes
Launch Page crashes if pipeline has dynamic parameters with only one option.  FIxed by adding a check and added a single disabled text input with the value.

## Related issue
N/A

## Checklist


ericenns (approved):
Everything looks to me and I verified by installing the species-abundance-plugin and only created 1 kraken database and 1 bracken database. As both of those are dynamic options pulled from galaxy.

PR #1146, opened 2021-12-10
Use forwarding rather than redirect for login page

innovate-invent:
## Description of changes
This sends unauthenticated users the login page in the response rather than sending a redirect.
Spring appears to ignore the tomcat configuration that handles headers from upstream proxies doing SSL offloading.
This causes a cross scheme redirect to be returned forcing users to use http over https.

Enabling relative redirects is the next best option over this PR:
https://github.com/OpenLiberty/open-liberty/issues/10000

I think I narrowed down where the cross scheme root redirect is occurring, but I need someone more familiar with tomcat and the spring framework to have a look at it and the information in this post:
https://stackoverflow.com/questions/3401113/spring-mvc-redirect-prefix-always-redirects-to-http-how-do-i-make-it-stay

Including the following in tomcats `server.xml:/Server/Service/Engine` does not resolve the issue:
```
<Valve className="org.apache.catalina.valves.RemoteIpValve"
    remoteIpHeader="x-forwarded-for"
    remoteIpProxiesHeader="x-forwarded-by"
    protocolHeader="x-forwarded-proto"
    protocolHeaderHttpsValue="https"
/>
```

## Checklist

### DIFF
--- a/src/main/java/ca/corefacility/bioinformatics/irida/ria/security/CredentialsExpiredAuthenticationFailureHandler.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/ria/security/CredentialsExpiredAuthenticationFailureHandler.java
@@ -33,14 +33,14 @@ import com.google.common.collect.ImmutableList;
  *
  */
 @Component
-public class CredentialsExpriredAuthenticationFailureHandler extends SimpleUrlAuthenticationFailureHandler {
-	private static final Logger logger = LoggerFactory.getLogger(CredentialsExpriredAuthenticationFailureHandler.class);
+public class CredentialsExpiredAuthenticationFailureHandler extends SimpleUrlAuthenticationFailureHandler {
+	private static final Logger logger = LoggerFactory.getLogger(CredentialsExpiredAuthenticationFailureHandler.class);
 
 	private final PasswordResetService resetService;
 	private final UserService userService;
 
 	@Autowired
-	public CredentialsExpriredAuthenticationFailureHandler(PasswordResetService resetService, UserService userService) {
+	public CredentialsExpiredAuthenticationFailureHandler(PasswordResetService resetService, UserService userService) {
 		this.resetService = resetService;
 		this.userService = userService;
 	}
--- a/src/main/java/ca/corefacility/bioinformatics/irida/ria/security/CredentialsExpiredAuthenticationFailureHandler.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/ria/security/CredentialsExpiredAuthenticationFailureHandler.java
@@ -43,6 +43,7 @@ public class CredentialsExpiredAuthenticationFailureHandler extends SimpleUrlAut
 	public CredentialsExpiredAuthenticationFailureHandler(PasswordResetService resetService, UserService userService) {
 		this.resetService = resetService;
 		this.userService = userService;
+		this.setUseForward(true);
 	}
 
 	/**
--- a/src/main/java/ca/corefacility/bioinformatics/irida/config/security/IridaWebSecurityConfig.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/config/security/IridaWebSecurityConfig.java
@@ -19,7 +19,7 @@ import org.springframework.web.servlet.LocaleResolver;
 
 import ca.corefacility.bioinformatics.irida.repositories.user.UserRepository;
 import ca.corefacility.bioinformatics.irida.ria.config.filters.SessionFilter;
-import ca.corefacility.bioinformatics.irida.ria.security.CredentialsExpriredAuthenticationFailureHandler;
+import ca.corefacility.bioinformatics.irida.ria.security.CredentialsExpiredAuthenticationFailureHandler;
 import ca.corefacility.bioinformatics.irida.ria.security.LoginSuccessHandler;
 
 /**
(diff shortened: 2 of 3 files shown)


ericenns:
I am unable to recreate your issue (i.e. meaning that I was able to get `RemoteIpValve` to successfully work.

We use Apache HTTPD as our revere ssl proxy in front of tomcat.

With the following Apache HTTPD config.

```
<VirtualHost *:443>
    ServerName              https://iridassl.local
    ErrorLog                /var/log/httpd/irida.ajp.error.log
    CustomLog               /var/log/httpd/irida.ajp.log combined
    ErrorDocument   503     /proxy-error/503.html

    # SSL Configs
    SSLEngine on
    SSLCertificateFile              /etc/ssl/certs/server.crt
    SSLCertificateChainFile /etc/ssl/certs/chainbundle.crt
    SSLCertificateKeyFile   /etc/ssl/certs/server.key

    <Proxy *>
        AddDefaultCharset Off
        Order deny,allow
        Allow from all
    </Proxy>

    ProxyPreserveHost       On
    ProxyPass                       /       http://iridassl.local:8080/irida
    ProxyPassReverse        /       http://iridassl.local:8080/irida

    RequestHeader set X-Forwarded-Proto https
    RequestHeader set X-Forwarded-Port 443

    # REST API:
    # IRIDA REST API (deployed to $TOMCAT_HOME/webapps/ as irida-web.war):
    <Location /irida>
        ProxyPass http://iridassldl.cscscience.ca:8080/irida connectiontimeout=7200 timeout=7200
    </Location>
</VirtualHost>

<VirtualHost *:80>
    ServerName              http://iridassl.local
    ErrorLog                /var/log/httpd/irida.ajp.error.log
    CustomLog               /var/log/httpd/irida.ajp.log combined
    ErrorDocument   503     /proxy-error/503.html

    <Proxy *>
        AddDefaultCharset Off
        Order deny,allow
        Allow from all
    </Proxy>

    # REST API:
    # IRIDA REST API (deployed to $TOMCAT_HOME/webapps/ as irida-web.war):
    <Location /irida>
        ProxyPass http://iridassldl.cscscience.ca:8080/irida connectiontimeout=7200 timeout=7200
    </Location>
</VirtualHost>
```

And no addition of `RemoteIpValve` to tomcat server.xml.

Going to `https://iridassl.local` redirects to `http://iridassl.local/irida/login`. (The error you describe)

Adding the following to tomcats `server.xml:/Server/Service/Engine/Host`

```
<Valve className="org.apache.catalina.valves.RemoteIpValve"
    remoteIpHeader="x-forwarded-for"
    protocolHeader="x-forwarded-proto" />
```

Going to `https://iridassl.local` now redirects to `https://iridassl.local/irida/login`.

Note: Apache HTTPD and Tomcat are on the same same host, and the `Host` section in tomcats `server.xml` looks like:

`<Host name="localhost" appBase="webapps" unpackWARs="true" autoDeploy="True">`

Also when I created the `RemoteIpValve` I had `internalProxies=127.0.0.1` and that caused the old behavior to happen (So internalProxies wasn't being matched as I suspect). Removing that line caused it to use tomcat's default which is `By default, 10/8, 192.168/16, 169.254/16, 127/8, 172.16/12, and ::1 are allowed.`.

innovate-invent:
I originally had the valve at server.xml:/Server/Service/Engine but moved it to server.xml:/Server/Service/Engine/Host despite the documentation saying that it works at either level. It still does not work and issues a redirect to http. I have internalProxies unset and checked that my load balancer subnet is 10.0.130.0/24 which should match the default regex. 

Are you aware of anyone testing IRIDA with a non-local, non-apache service in-front of tomcat?

src/main/java/ca/corefacility/bioinformatics/irida/ria/security/CredentialsExpiredAuthenticationFailureHandler.java
```diff
@@ -33,16 +33,17 @@
...
 
 	private final PasswordResetService resetService;
 	private final UserService userService;
 
 	@Autowired
-	public CredentialsExpriredAuthenticationFailureHandler(PasswordResetService resetService, UserService userService) {
+	public CredentialsExpiredAuthenticationFailureHandler(PasswordResetService resetService, UserService userService) {
 		this.resetService = resetService;
 		this.userService = userService;
+		this.setUseForward(true);
```
innovate-invent, line 46: If redirects want to be kept, then we could change this to `this.getRedirectStrategy().setContextRelative(true)`

ericenns:
> I originally had the valve at server.xml:/Server/Service/Engine but moved it to server.xml:/Server/Service/Engine/Host despite the documentation saying that it works at either level. It still does not work and issues a redirect to http. I have internalProxies unset and checked that my load balancer subnet is 10.0.130.0/24 which should match the default regex. 
> 
> Are you aware of anyone testing IRIDA with a non-local, non-apache service in-front of tomcat?

Definitely not matching the remoteIpValve for your deployment. I suspect that it is using a different IP then expected.

I found this special version of remoteIpValve for AWS. https://github.com/Collaborne/tomcat-remoteip-aws-valve you could look at it get a list of IP ranges to add yourself.

One other side effect that using remoteIpValve fixed for us was the REST API resource link responses. Without it when connecting to an https rest endpoint the responses would include URIs that were http. After adding in the valve the URIs now correctly start with https.

ericenns:
I suggest making a discussion under the discussions tab of the repo vs continuing in this closed PR.

PR #1150, opened 2021-12-21
Spring boot

### LINKED ISSUE #852: IRIDA fails to connect to MySQL 8
## Describe the bug
IRIDA fails to connect to MySQL 8

In our [dockerized](https://github.com/COMBAT-TB/irida-galaxy-deploy) setup, we initially had IRIDA combined with MySQL 8. At startup, the IRIDA Servlet failed to connect to the database and thus failed to startup. Replacing MySQL with MariaDB solved this problem.

## Expected behaviour
According to the documentation, IRIDA works with both MySQL and MariaDB so this was unexpected.

## Additional context
Further investigation revealed that in MySQL 8 they switched to [caching_sha2_password](https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html) as the default authentication plugin. Apparently this plugin is only [supported](https://stackoverflow.com/questions/50387952/how-to-resolve-unable-to-load-authentication-plugin-caching-sha2-password-issu) in versions 5.1.46 and later of the MySQL Connector for Java. IRIDA is currently using version [5.1.31](https://github.com/phac-nml/irida/blob/bda8c844a41d3516a9cbe45bb39affd75b66a8a8/pom.xml#L1118) of that connector.

Testing with a very simple JDBC script verified that upgrading to version 5.1.46 of the connector resolves the problem, however, this revealed a [SSL related error](https://stackoverflow.com/questions/34189756/warning-about-ssl-connection-when-connecting-to-mysql-database) which required adding a `useSSL=false` parameter to the connect string. The need for this parameter should be noted in the IRIDA documentation and example config file.

ericenns:
## Description of changes

Converted the application from pure Spring to Spring Boot. A number of dependencies were updated, major ones being:
* Spring 5.2.2.RELEASE -> 5.3.15
* Hibernate 5.4.7.Final -> 5.6.4.Final
* Liquibase 3.5.1 -> 4.5.0
* mysql-connector-java 5.1.31 -> 8.0.28
* thymeleaf 3.0.9.RELEASE -> 3.0.14.RELEASE
* mockito 1.10.19 -> 4.0.0
* rest-assured 2.3.4 -> 4.4.0

Logging was switched from Log4J2 to Logback.

Development web server was changed from Jetty to Tomcat.

Liquibase changsets were updated to support mariadb:10.4 and mysql:8 (note: Did not need to add `validChecksum` since liquibase was updated and the checksum alg changed so liquibase will just updated checksums in DATABASECHANGELOG and trust that the changesets are the same.).

## Related issue

Fixes #852

## Checklist


src/test/java/ca/corefacility/bioinformatics/irida/ria/unit/web/analysis/AnalysesListingServiceTest.java
```diff
@@ -56,8 +58,8 @@ public void testGetPagedSubmissionsForAdmin() throws IridaWorkflowNotFoundExcept
 				ImmutableMap.of());
 
 		when(analysisSubmissionService
-				.listAllSubmissions(eq(searchValue), any(String.class), eq(null), Matchers.any(),
-						Matchers.any())).thenReturn(AnalysesDataFactory.getPagedAnalysisSubmissions());
+				.listAllSubmissions(eq(searchValue), isNull(), isNull(), isNull(),
```
deepsidhu85, line 61: For the any(String.class) here and in tests below why are we not checking for isNull()?

src/test/java/ca/corefacility/bioinformatics/irida/ria/unit/web/oauth/RemoteAPIAjaxControllerTest.java
```diff
@@ -123,7 +123,7 @@ public Pageable previousPageable() {
 		};
 
 		when(remoteAPIService.search(any(), anyInt(), anyInt(), any(Sort.Direction.class),
-				any(String[].class))).thenReturn(remoteAPIPage);
+				any(String.class))).thenReturn(remoteAPIPage);
```
deepsidhu85, line 126: Is there a reason for this change from `String[].class` to `String.class`?

src/test/java/ca/corefacility/bioinformatics/irida/ria/unit/web/projects/ProjectSamplesControllerTest.java
```diff
@@ -306,7 +309,7 @@ public void testGetAjaxProjectSampleModels() {
 		));
 
 		when(sampleService
-				.getFilteredSamplesForProjects(any(List.class), any(List.class), any(String.class), any(String.class), any(String.class), any(Date.class), any(Date.class),
```
deepsidhu85, line 309: Same question as for another test above about why are we now checking for `isNull()` instead of `any(String.class)`. For this test why are we also checking for `isNull() in place of these `any(Date.class)` `any(Date.class)`

src/test/java/ca/corefacility/bioinformatics/irida/service/impl/unit/user/UserServiceImplTest.java
```diff
@@ -137,7 +135,7 @@ public void testReadNewAnnouncements() {
 		Announcement second_announcement = new Announcement("test 2", "this is also a test message", false, u);
 		List<Announcement> unreadAnnouncements = Lists.newArrayList(first_announcement, second_announcement);
 
-		when(announcementUserJoinRepository.getAnnouncementsUnreadByUser(any(User.class))).thenReturn(unreadAnnouncements);
```
deepsidhu85, line 140: Another one, any reason for checking for `isNull()` in place of `any(User.class)`?

deepsidhu85 (changes requested):
This is looking really good. Just a few comments/changes/questions below. Other than that I just have to test it out thoroughly

pom.xml
```diff
@@ -362,16 +263,6 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
-		<dependency>
-			<groupId>com.fasterxml.jackson.core</groupId>
```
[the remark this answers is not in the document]
ericenns: `jackson` is pulled in by `jena-tdb`. To see this you can run `mvn dependency:tree` and then look at the tree based output, i.e:
```
[INFO] +- org.apache.jena:jena-tdb:jar:4.3.2:compile
[INFO] |  \- org.apache.jena:jena-arq:jar:4.3.2:compile
[INFO] |     +- org.apache.jena:jena-core:jar:4.3.2:compile
[INFO] |     |  +- org.apache.jena:jena-base:jar:4.3.2:compile
[INFO] |     |  |  +- org.apache.jena:jena-shaded-guava:jar:4.3.2:compile
[INFO] |     |  |  \- com.github.andrewoma.dexx:collection:jar:0.7:compile
[INFO] |     |  \- org.apache.jena:jena-iri:jar:4.3.2:compile
[INFO] |     +- com.github.jsonld-java:jsonld-java:jar:0.13.3:compile
[INFO] |     +- com.apicatalog:titanium-json-ld:jar:1.1.0:compile
[INFO] |     +- org.glassfish:jakarta.json:jar:2.0.1:compile
[INFO] |     +- com.fasterxml.jackson.core:jackson-core:jar:2.13.1:compile
[INFO] |     +- com.fasterxml.jackson.core:jackson-databind:jar:2.13.1:compile
[INFO] |     |  \- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.1:compile
[INFO] |     +- org.apache.httpcomponents:httpclient-cache:jar:4.5.13:compile
[INFO] |     +- com.google.protobuf:protobuf-java:jar:3.17.3:compile
[INFO] |     +- org.apache.thrift:libthrift:jar:0.15.0:compile
[INFO] |     |  \- javax.annotation:javax.annotation-api:jar:1.3.2:compile
[INFO] |     \- org.apache.commons:commons-lang3:jar:3.12.0:compile
```

pom.xml
```diff
@@ -578,23 +395,14 @@
...
-			<version>${jsonpath.version}</version>
-			<scope>test</scope>
-		</dependency>
-
 		<!-- Logging -->
 		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-api</artifactId>
-			<version>${slf4j.version}</version>
+			<groupId>org.apache.logging.log4j</groupId>
```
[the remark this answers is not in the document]
ericenns: `log4j` did have a vulnerability but this package is just the binding between it and `slf4j` which is required by `apache.poi`. In the `properties` section, I have a `log4j2.version` defined which is set to 2.17.1 which addressed the vulnerability.

src/main/java/ca/corefacility/bioinformatics/irida/config/environment/IridaEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+			.put("jdbc.pool.removeAbandonedTimeout", "spring.datasource.dcp2.remove-abandoned-timeout")
+			.put("jdbc.pool.maxIdle", "spring.datasource.dcp2.max-idle")
+			.build();
+
+		MutablePropertySources propertySources = env.getPropertySources();
+
+		Properties properties = new Properties();
+		for (Map.Entry<String, String> entry : deprecatedPropertiesMap.entrySet()) {
+			if (env.containsProperty(entry.getKey())) {
+				this.logger.error("Translating deprecated property " + entry.getKey() + " to " + entry.getValue());
```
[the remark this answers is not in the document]
ericenns: I had set this as `ERROR` because the properties are deprecated and I wanted to make sure admins that had there logging level set to `ERROR` (which I think is default, might be wrong) would see it.

src/main/java/ca/corefacility/bioinformatics/irida/config/data/IridaApiJdbcDataSourceConfig.java
```diff
@@ -43,17 +51,17 @@ public ApplicationContextAwareSpringLiquibase(final ApplicationContext applicati
 		}
 
 		@Override
-		protected SpringResourceOpener createResourceOpener() {
-			return new ApplicationContextSpringResourceOpener(getChangeLog());
+		protected SpringResourceAccessor createResourceOpener() {
+			return new ApplicationContextSpringResourceOpener(resourceLoader);
```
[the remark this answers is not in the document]
ericenns: `resourceLoader` is an attribute of `SpringLiquibase`, so not sure how to comment that.

src/test/java/ca/corefacility/bioinformatics/irida/ria/unit/web/analysis/AnalysesListingServiceTest.java
ericenns: Mockito 1.x which we were on would accept `null` values for `any()` even if a class if expected.

src/test/java/ca/corefacility/bioinformatics/irida/ria/unit/web/oauth/RemoteAPIAjaxControllerTest.java
ericenns: `String[].class` appeared to have been an error which was allowed by Mockito 1.x

src/test/java/ca/corefacility/bioinformatics/irida/ria/unit/web/projects/ProjectSamplesControllerTest.java
ericenns: Same as above comment:

Mockito 1.x which we were on would accept null values for any() even if a class is expected.

src/test/java/ca/corefacility/bioinformatics/irida/service/impl/unit/user/UserServiceImplTest.java
ericenns: Because the call to the function actually passed null.

Mockito 1.x which we were on would accept null values for any() even if a class if expected.

src/main/java/ca/corefacility/bioinformatics/irida/config/environment/IridaEnvironmentPostProcessor.java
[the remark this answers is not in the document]
ericenns: Addressed in c3f20422

src/test/java/ca/corefacility/bioinformatics/irida/config/data/IridaDbUnitConfig.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.github.springtestdbunit.bean.DatabaseConfigBean;
+import com.github.springtestdbunit.bean.DatabaseDataSourceConnectionFactoryBean;
+
+import org.dbunit.ext.mysql.MySqlDataTypeFactory;
+import org.dbunit.ext.mysql.MySqlMetadataHandler;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
```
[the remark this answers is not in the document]
ericenns: Addressed in c3f20422

src/main/java/ca/corefacility/bioinformatics/irida/config/data/IridaApiJdbcDataSourceConfig.java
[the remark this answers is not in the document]
ericenns: Addressed in c3f20422

deepsidhu85 (approved):
Looks good to me and works great. Thanks for making those few changes! @apetkau did you want to go through it as well and see if I missed anything?

src/main/resources/ca/corefacility/bioinformatics/irida/database/changesets/0.20.0/pipeline-metadata-entry.xml
```diff
@@ -7,9 +7,6 @@
 
     <changeSet id="pipeline-metadata-entry" author="tom">
 
-        <validCheckSum>7:04d990c0430b5fd7bd63f5bfc462584b</validCheckSum>
-        <validCheckSum>7:fe424670afa75882a10c9178073eecc1</validCheckSum>
```
JeffreyThiessen, line 11: Why are the checksums being removed? Did this move somewhere else? I noticed checksums being removed on other files too.

JeffreyThiessen (commented):
Looks good. Just have one comment/question below.

src/main/resources/ca/corefacility/bioinformatics/irida/database/changesets/0.20.0/pipeline-metadata-entry.xml
ericenns: The `validCheckSum` tags were removed as they are no longer required. The previous version of liquibase that we were on had version 7 of the checksum alg (As denoted by the first number before the colon in `7:04d990c0430b5fd7bd63f5bfc462584b`), and the new version has version 8 of the checksum alg.

See here for the block of code where liquibase will check if the checksums are compatible and if not it will nullify them and then set them to the newly calculated value. (https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java#L250)

JeffreyThiessen (approved):

PR #1164, opened 2022-01-07
Zipped html support

### LINKED ISSUE #162: blend4j: HistoriesClient#downloadDataset mangles HTML file contents
Described here: https://github.com/galaxyproject/blend4j/issues/38

I'm not sure if it only affects HTML format files or other file types as well. 

It might be a good idea to change the code for downloading output files from Galaxy to use `org.apache.commons.io.FileUtils#copyURLToFile` instead of `HistoriesClient#downloadDataset`.

Maybe change http://gitlab-irida.corefacility.ca/irida/irida/blob/development/src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/AnalysisWorkspaceServiceGalaxy.java#L155 to 

```java
final URL url = new URL(dataset.getFullDownloadUrl());
Path outputFile = outputDirectory.resolve(fileName);
FileUtils.copyURLToFile(url, outputFile.toFile());
```

Imported from GitLab issue #650.  Originally posted on 2018/05/29 05:33PM\Posted by Peter Kruczkiewicz

### LINKED ISSUE #1057: Deal with Galaxy zipped HTML
When a Galaxy tool outputs HTML it is downloaded from Galaxy as a Zipped HTML. If IRIDA tries to display this (if the dataset is named something .html) it gets displayed as binary data.

One workaround at present is to change the datatype on the Galaxy workflow side (to something like txt) so that it gets downloaded as a text file.

Ideally IRIDA would be able to pick up the "zipped HTML", unpack the HTML from the zip and display that.

This would be interact with the HTML viewer (https://github.com/phac-nml/irida/pull/849/files#) and the [AnalysisController](https://github.com/phac-nml/irida/blob/development/src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java).

## Additional information
A known limitation is that the current IRIDA HTML viewer expects a single web page. If all the assets of a web page are loaded into this HTML it can get large and unwieldy.

ericenns:
## Description of changes

Add in better handling for zipped html outputs.

## Related issue

Fixes #1057
Fixes #162

## Checklist


src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/AnalysisWorkspaceServiceGalaxy.java
```diff
@@ -138,8 +141,16 @@ private AnalysisOutputFile buildOutputFile(String analysisId, String labelPrefix
 		String datasetId = dataset.getId();
 		String fileName = dataset.getName();
 
+		Path tmpOutputFile = Files.createTempFile(dataset.getName(), null);
+		tmpOutputFile.toFile().deleteOnExit();
+
+		galaxyHistoriesService.downloadDatasetTo(analysisId, datasetId, tmpOutputFile);
 		Path outputFile = outputDirectory.resolve(fileName);
-		galaxyHistoriesService.downloadDatasetTo(analysisId, datasetId, outputFile);
+		if (FileUtilities.isZippedFile(tmpOutputFile)) {
```
apetkau, line 149: So I'm wondering if we actually need to rename output files to add `.zip` here? Could we not look for an `.html` extension when loading the file in the **Analysis** page and open it up to check if it's a zip file (without needing it to end in `.html.zip`)? I guess it messes up with downloading the data (since you'll download a `.html` file even though it's a `.zip` file, but maybe we could modify the code for downloading output files to append a `.zip` there?

The reason I am thinking it would be nice to avoid re-naming files when saving, is that then this code to extract HTML files from a zip `.html` archive will work with all existing analysis submissions. In particular, the **Assembly Annotation Collection** pipeline already has QUAST results in a `.html` zip archive and if we remove any re-naming of output files from this code we could also just automatically display QUAST results from previous runs.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
```diff
@@ -200,23 +206,57 @@ public void getHtmlOutputForSubmission(@PathVariable Long submissionId, @Request
...
-			OutputStream outputStream = response.getOutputStream();
-			/*
-			Write the htmlNotFound message to the outputstream. We do this
-			so that the page doesn't error and will instead display the
-			message.
-			 */
-			outputStream.write(htmlOutputNotFound.getBytes(StandardCharsets.UTF_8));
-			outputStream.flush();
-			outputStream.close();
+		if (FileUtilities.isZippedFile(outputFile.getFile())) {
```
apetkau, line 209: I'm wondering if you could add in a check for `.zip` file extension in addition to using `isZippedFile()` here? Or maybe check for a `.html` extension (if we chose to not rename `.html` to `.html.zip`)?

The reason being is that some files (like Excel `xlsx`) are actually zip files with a renamed extension and we don't want to open them up and look for HTML files.

apetkau (changes requested):
This looks amazing to me @ericenns . Thanks so much. It looks like it works:

image

Not only that, but I **think** we can get this to also work to load up HTML files from existing pipelines that were previously not displayed. In particular the QUAST results from the **Assembly Annotation Collection** pipeline:

image

However, I think to get this to work properly there are some changes that need to be made (comments given below).

src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/AnalysisWorkspaceServiceGalaxy.java
```diff
@@ -138,8 +141,16 @@ private AnalysisOutputFile buildOutputFile(String analysisId, String labelPrefix
...
+			outputFile = outputDirectory.resolve(fileName + ".zip");
```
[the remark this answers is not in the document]
ericenns: I reverted this file in 4c57b66

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
ericenns: This function is now just being call inside of `FileUtilities.getFileExt` when it determines that ext is `html` from regex. Then if it is zipped html `FileUtilities.getFileExt` will return `html-zip`.

src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/AnalysisWorkspaceServiceGalaxy.java
ericenns: I addressed that in 4c57b66 and in 951c063 I changed `getLabel` in `AnalysisOutputFile` to append zipped html files with `.zip` so that downloading works as expected.

src/main/java/ca/corefacility/bioinformatics/irida/ria/utilities/FileUtilities.java
```diff
@@ -236,20 +234,44 @@ private static String formatName(String name) {
...
-		Matcher matcher = regexExt.matcher(filename);
+	public static String getFileExt(Path filepath) {
+		Matcher matcher = regexExt.matcher(filepath.getFileName().toString());
 		String ext = "";
 		if (matcher.matches()) {
-			ext = matcher.group(1);
+			ext = matcher.group(1).toLowerCase();
 		}
-		return ext.toLowerCase();
+		if (ext.equals("html")) {
```
apetkau, line 250: Do you think you could maybe add another condition in here `ext.equals("html.zip")`? That way we can also display the `quast.html.zip` output from previous run results of the **Assembly Annotation Collection** pipeline (though there are multiple `html` files in the zip archive in this case and only one will be chosen, though I think that's alright for now).

Maybe going too far outside of the scope of what we want to accomplish though, since we mainly just want to support a single zipped HTML file.

src/main/java/ca/corefacility/bioinformatics/irida/service/analysis/workspace/galaxy/AnalysisWorkspaceServiceGalaxy.java
apetkau: Awesome. Thanks.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/analysis/AnalysisController.java
apetkau: Awesome. Sounds great.

apetkau (changes requested):
Thanks so much for making these changes Eric. It works great and now doesn't require any special changes to the output file names.

One other comment is given below and additionally I'm wondering if you could add a message to `CHANGELOG.md`?

ericenns:
> Thanks so much for making these changes Eric. It works great and now doesn't require any special changes to the output file names.
> 
> One other comment is given below and additionally I'm wondering if you could add a message to `CHANGELOG.md`?

Yes I can update the `CHANGELOG.md`, I just wanted to make sure that my solution was even on the right page before updating it. @deepsidhu85 and @joshsadam have also been looking at it and giving feedback along the way and it appears we are all happy with this solution.

src/main/java/ca/corefacility/bioinformatics/irida/ria/utilities/FileUtilities.java
apetkau: Awesome. It works. Thanks so much:

image

apetkau (approved):
Everything looks great to me. Thanks so much for making those changes.

deepsidhu85 (commented):
Everything looks good to me as well. Thanks for going through it with me yesterday and showing me how it works. Once you add the documentation changes I will merge it. Also thanks @apetkau for looking through it and testing it out

apetkau:
Thanks @deepsidhu85 

Also, thanks @pvanheus for your other PR (https://github.com/phac-nml/irida/pull/1164) which helped us out with this one. I hope the changes in this PR all work for you.

ericenns:
> Everything looks good to me as well. Thanks for going through it with me yesterday and showing me how it works. Once you add the documentation changes I will merge it. Also thanks @apetkau for looking through it and testing it out

I added in some documentation for the html output preview, let me know if you want more detail or if the info I provided is sufficient.

PR #1166, opened 2022-01-12
Metadata field security no projects alert

joshsadam:
## Description of changes

Added an `Alert` that tells the user if there are no projects available to share with.  This could be do to the fact that they only have one project or that they are not a manager on any other projects.

## Related issue

N/A

## Checklist

*   [ ] ~CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change~. Will be added to final PR
*   [ ] ~Tests added (or description of how to test) for any new features~ Will be added to the final PR
*   [ ] ~User documentation updated for UI or technical changes~.Will be added to final PR


deepsidhu85 (approved):
This works great and looks good. Will merge once we get dev merged up into our metadata field security stuff

src/main/webapp/resources/js/pages/projects/share/ShareProject.jsx
```diff
@@ -1,44 +1,28 @@
 import { Select, Space, Typography } from "antd";
 import React from "react";
 import { useDispatch, useSelector } from "react-redux";
-import {
-  useGetPotentialProjectsToShareToQuery
-} from "../../../apis/projects/projects";
+import { useGetPotentialProjectsToShareToQuery } from "../../../apis/projects/projects";
```
deepsidhu85, line 4: Don't think this is being used and can be removed?

src/main/webapp/resources/js/pages/projects/share/ShareProject.jsx
```diff
@@ -1,44 +1,28 @@
...
 import { setProject } from "./shareSlice";
 
 /**
  * React component for selecting the project to share a sample with.
+ * @param {list} projects - list of projects that the user is a manager on
  * @returns {JSX.Element}
  * @constructor
  */
-export function ShareProject() {
+export function ShareProject({ projects }) {
```
deepsidhu85, line 13: Can we get the projects from the state here instead of passing it in?

src/main/webapp/resources/js/pages/projects/share/index.js
```diff
@@ -130,61 +148,82 @@ function ShareApp() {
...
-                        justifyContent: "space-between",
-                      }}
+        <Skeleton loading={projectsLoading} active>
+          <Card>
+            <PageHeader
+              ghost={false}
+              title={i18n("ShareSamples.title")}
+              onBack={goToPrevious}
+            >
+              {projects && projects.length === 0 ? (
```
deepsidhu85, line 158: I think we can change this to `projects?.length === 0`

deepsidhu85 (changes requested):
Looks great and working as expected. Just a few small comments below otherwise good to go

src/main/webapp/resources/js/pages/projects/share/ShareProject.jsx
joshsadam: Removed in 6d21c776a1f9698293d9d200534ff09576a74775

src/main/webapp/resources/js/pages/projects/share/ShareProject.jsx
joshsadam: Ignoring :)

src/main/webapp/resources/js/pages/projects/share/index.js
joshsadam: I like this, nice find :) updated in 483fa813c506a861710628ef8da40288bc501821

deepsidhu85 (approved):

Issue #1108, opened 2021-10-12
Update reach-router to react -router


joshsadam:
## What needs changed?

Update pages using `reach-router` to `react-router` v6.  Why?  They are sibling projects, very similar API, but future development will be on `react-router`, since there are only a few pages to update, this should be done now.

## How does it work now?

`reach-router` works now, this is a future preparedness update

## How should it work?

How do you think this component should work?

## Additional information

ericenns:
Fixed in 22.01

PR #1171, opened 2022-01-19
Remove @reach/router from pages that only us it for getting the project identifier

joshsadam:
## Description of changes
Removes @reach/router from pages that only used it to get the project id from the URL.  Instead, capture with a regular expression.  Created a function to handle this in url-utilities.

## Related issue
#1108 

## Checklist


src/test/java/ca/corefacility/bioinformatics/irida/ria/integration/projects/ProjectRemoteSettingsIT.java
```diff
@@ -53,10 +53,6 @@ public void testRemoteProjectSettings() {
 		assertFalse("URL should not be empty", url.isEmpty());
 		page.submitProject();
 
-		ProjectDetailsPage projectDetailsPage = ProjectDetailsPage.initElements(driver());
```
ericenns, line 56: Does `submitProject` have a changed behaviour? It looks like previously according to this test it went to the `ProjectDetailsPage` and now it goes to the `ProjectRemoteSettingsPage`?

ericenns (approved):
👍 everything works! thanks!

PR #1187, opened 2022-02-04
Hotfix/correct rest api date fields

ericenns:
## Description of changes
In 22.01 I incorrectly changed the response format of the REST API date fields to textual string instead of epoch ms. This reverts back to the old behavior and updates the UPGRADING.md to mention issues we faced when deploying 22.01.

## Related issue

## Checklist

### DIFF
--- a/src/main/java/ca/corefacility/bioinformatics/irida/web/controller/api/json/TimestampJson.java
+++ b/src/main/java/ca/corefacility/bioinformatics/irida/web/controller/api/json/TimestampJson.java
@@ -1,0 +1,38 @@
+package ca.corefacility.bioinformatics.irida.web.controller.api.json;
+
+import java.io.IOException;
+import java.util.Calendar;
+import java.util.Date;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import org.apache.commons.lang3.time.DateUtils;
+
+/**
+ * Serialization class for java Date objects. We use DATETIME types for storing
+ * our timestamps in mysql. DATETIME only stores at seconds precision by
+ * default. In the REST api when we create an object the timestamps will have
+ * milliseconds precision but on subsequent fetchs the timestamps will only have
+ * seconds precision. This enforces the timestamp fields to only ever have
+ * seconds precision.
+ */
+public class TimestampJson {
+
+	/**
+	 * Default serializer for {@link Date} objects.
+	 */
+	public static class TimestampSerializer extends StdSerializer<Date> {
+
+		public TimestampSerializer() {
+			super(Date.class);
+		}
+
+		@Override
+		public void serialize(Date value, JsonGenerator gen, SerializerProvider provider) throws IOException {
+			gen.writeNumber(DateUtils.truncate(value, Calendar.SECOND).getTime());
+		}
+
+	}
+}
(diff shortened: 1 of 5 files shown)


src/main/java/ca/corefacility/bioinformatics/irida/web/controller/api/json/TimestampJson.java
```diff
@@ -0,0 +1,38 @@
...
+	 */
+	public static class TimestampSerializer extends StdSerializer<Date> {
+
+		public TimestampSerializer() {
+			super(Date.class);
+		}
+
+		@Override
+		public void serialize(Date value, JsonGenerator gen, SerializerProvider provider) throws IOException {
+			gen.writeNumber(DateUtils.truncate(value, Calendar.SECOND).getTime());
```
apetkau, line 34: Just to make sure I understand, we are truncating (instead of rounding) because the milliseconds are being truncated in the database (and not rounded)?

apetkau (commented):
This looks awesome. Thanks so much for the fix.

I tested out launching a new instance of IRIDA and running pipelines/getting results using `irida-staramr-results` and `irida-sistr-results`. Both of these now work properly, so the API is returning the same values as before.

One other general comment I had is that I noticed when testing in my log files I am getting a lot of `Hibernate` SQL statements in the logs:

```
Hibernate: select analysissu0_.id as id2_6_, analysissu0_.analysis_description as analysis3_6_, analysissu0_.created_date as created_4_6_, analysissu0_.email_pipeline_result_completed as email_pi5_6_, analysissu0_.email_pipeline_result_error as email_pi6_6_, analysissu0_.modified_date as modified7_6_, analysissu0_.name as name8_6_, analysissu0_.named_parameters_id as named_p20_6_, analysissu0_.reference_file_id as referen21_6_, analysissu0_.submitter as submitt22_6_, analysissu0_.update_samples as update_s9_6_, analysissu0_.workflow_id as workflo10_6_, analysissu0_.analysis_id as analysi23_6_, analysissu0_.analysis_cleaned_state as analysi11_6_, analysissu0_.analysis_state as analysi12_6_, analysissu0_.automated as automat13_6_, analysissu0_.priority as priorit14_6_, analysissu0_.remote_analysis_id as remote_15_6_, analysissu0_.remote_input_data_id as remote_16_6_, analysissu0_.remote_workflow_id as remote_17_6_ from analysis_submission analysissu0_ where analysissu0_.DTYPE='AnalysisSubmission' and analysissu0_.analysis_state=?
...
```

Do you think I have something misconfigured here, or is the default settings for logging set to log SQL statements?

src/main/java/ca/corefacility/bioinformatics/irida/web/controller/api/json/TimestampJson.java
ericenns: From my testing it was failing randomly when using `round` and when using `truncate` it never failed. So by that I assume it is truncating.

apetkau: Yes, I agree. I observed the same thing (if we use the wrong method then some of the tests should fail 50% of the time, but I tried them out a bunch of times and they keep on passing, which means it must be `truncated` and not `rounded`).

apetkau (approved):
This looks great.

For the printing of SQL statements in the logs, this was my mistake. I somehow had the following set in my `/etc/irida/irida.conf`:

```
hibernate.show_sql=true
```

Disabling that means no more SQL gets printed.

PR #1208, opened 2022-02-23
hotfix: Fixed syncing remote projects from version 20.09 and older.

### LINKED ISSUE #1206: 22.01 and greater cannot sync projects with instances below version 21.01
## Describe the bug
Null pointer exception occurs in ProjectRemoteRepositoryImpl.java:66 when trying access project/hash url as the if statement above does not trigger as it is null checking an Optional, which would never be null. Instead the Optional should be checked to be not empty instead.

ericenns:
## Description of changes
Fixed issue with syncing remote projects from version 20.09 and older.

## Related issue
Fixes #1206 

## Checklist


deepsidhu85 (approved):
Looks good. Thanks for fixing it and adding a test as well.

Issue #1066, opened 2021-07-30
Order statistics charts by date/time instead of string


apetkau:
## Describe the bug
It looks like the order of some of the columns in the statistics charts are by the string label. For charts that show information related to the date/time it would be nice if the labels were ordered by date/time instead.

![image](https://user-images.githubusercontent.com/200517/127678166-fa476dfe-9656-4576-b72f-bbf356e00684.png)

For example, here you can see that the column for July 2020 and 2021 (`07/20`, `07/21`) are next to each other, instead of grouped by the year.

## Steps to reproduce the problem
1. Open up the **Admin** panel in IRIDA
2. Look at some of the charts (e.g,. Analyses)

## Expected behaviour
I expected the order of the labels to be by date/time.

## Additional context
None.

apetkau:
Note if you encode the date as `YYYY/MM` then ordering by string should be the same as ordering by calendar date (though then you can't change the labels for dates later on to e.g, `July 2021` and so on).

ksierks:
Fixed in  #1221.

PR #1221, opened 2022-03-21
Sorting columns in admin statistic charts

ksierks:
## Description of changes
Sorted columns by date and changed daily format from "03/09" to "Mar 9" and monthly format from "07/21" to "Jan 2021". 
Note: Both daily and monthly formats start with month, which could be confusing when switching between time periods.
ex.) 03/21 could read as March 21 or March 2021.

image

image

## Related issue
Fixes issue #1066.

## Checklist


deepsidhu85 (changes requested):
This is looking great. Just a few changes. Let's remove the values that appear on top of the bars for either the Bar or Column chart types (if someone is viewing data for say the month then there could be 31 bars, and if a large number or projects, samples, or analyses were created on a certain day then the display could get messy. The values appear in the popover when we hover over a bar anyways so it's ok. I added a picture below in a comment of what it looks like with 31 bars.

Also, let's see if we can add some padding or something between the chart and the legend

deepsidhu85:
adminstats

src/main/webapp/resources/js/pages/admin/chart-config.js
```diff
@@ -41,15 +48,10 @@ export function getChartConfiguration(chartType, data) {
...
-    label: {
-      visible: Boolean(data),
-      position: "middle",
-      adjustColor: true,
-      style: { fill: "#0D0E68", fontSize: 12, fontWeight: 600, opacity: 0.3 },
-    },
     legend: {
       visible: Boolean(data),
       position: "bottom",
+      offsetY: 10
```
deepsidhu85, line 54: Can I get you to add `maxWidth: window.innerWidth - 600` in the legend section as well. That way the legend doesn't extend from the chart margin on the left to the right

deepsidhu85 (changes requested):
One last thing to add and it's good to go!

src/main/webapp/resources/js/pages/admin/chart-config.js
ksierks: Please see 8c5ca22cc7d591a5d0a7bb21f6559dba0b638564.

deepsidhu85 (approved):

PR #1235, opened 2022-04-01
Allow arbitrary text for organism

### LINKED ISSUE #1121: Cannot add organism to project not in list
## Describe the bug

It's not possible to assign a project an organism not in the list of existing names. This was possible previously:

image

1. Create a new project
2. Type in a new name for an organism
3. Continue with creation of the project
4. If the organism name is not found in the pre-defined list, it will be set to **blank**

## Expected behaviour
I expected to be able to set organism names outside of the list of approved names.

## Additional context
None.

joshsadam:
## Description of changes
All the user to enter a value that is not included in the taxonomy into the organism field when creating a project.

## Related issue
Fixes #1121

## Checklist

### DIFF
--- a/src/main/webapp/resources/js/components/ontology/OntologySelect.jsx
+++ b/src/main/webapp/resources/js/components/ontology/OntologySelect.jsx
@@ -1,9 +1,9 @@
-import { Select } from "antd";
+import { AutoComplete } from "antd";
 import React, { useEffect, useRef, useState } from "react";
 import { searchOntology } from "../../apis/ontology/taxonomy";
 import { useDebounce } from "../../hooks";
 
-const { Option } = Select;
+const { Option } = AutoComplete;
 
 /**
  * Component to render a select input to search for a term in an ontology.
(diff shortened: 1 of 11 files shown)


src/main/webapp/resources/js/components/ontology/OntologySelect.jsx
```diff
@@ -1,9 +1,9 @@
...
+import { AutoComplete } from "antd";
 import React, { useEffect, useRef, useState } from "react";
 import { searchOntology } from "../../apis/ontology/taxonomy";
 import { useDebounce } from "../../hooks";
 
-const { Option } = Select;
+const { Option } = AutoComplete;
 
 /**
  * Component to render a select input to search for a term in an ontology.
```
ericenns, line 9: Need to update comment here, as this is no longer a select input.

ericenns (changes requested):
Looks good, but could you please rename `OntologySelect` to `OntologyInput` and remove references to `select` within the file (i.e. comments).

ericenns (approved):
👍 As long as the tests pass, I'm happy with this. Thanks for fixing this issue.

PR #1272, opened 2022-05-13
Clarify messages related to deleting data from IRIDA

### LINKED ISSUE #1126: Make "Remove/Delete" clearer so it's obvious that it doesn't remove data from the filesystem
Right now we have options to delete a project, delete samples from a project, or delete sequence reads from a sample. These don't actually remove sequence data from the filesystem, just unlink from being visible to a user.

However, some of the descriptive text is either lacking or can be misleading. The term "Delete" or "Remove" could mislead people into thinking it removes sequence data. It would be very useful to either rename "Delete/Remove" to something else "Unlink?" or add additional descriptive text so that it's clear that deleting data is not removing the data from the filesystem.

Any of the places where we've implemented a "Delete/Remove" (e.g., delete a project, a sample, or sequence reads).

## Additional information

1. Delete a project

   image

2. Delete a sample from a project

   image

3. Delete sequence reads from a sample

   image

apetkau:
## Description of changes

Updates messages in UI when deleting **Projects**, **Samples**, **Sequence files**, and **Analyses** to clarify what it means to delete. Specifically, that underlying data is not removed from IRIDA, but simply made inaccessible (e.g., sequence data is not deleted, just removed from the UI).

## Related issue

Fixes #1126 

## Checklist


ericenns (approved):
Looks great, thanks for fixing this issue!

PR #1327, opened 2022-06-23
WIP: Sample details refactor/ default assembly

deepsidhu85:
## Description of changes

Added functionality to allow setting of a default genome assembly for a sample

## Related issue

## Checklist


src/main/resources/ca/corefacility/bioinformatics/irida/database/changesets/RENAME_ME_WHEN_READY/all-changes.xml
```diff
@@ -1,8 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-
-<databaseChangeLog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
joshsadam, line 3: Why is the outer folder named the way it is?

src/main/webapp/resources/js/components/samples/components/GenomeAssemblyList.jsx
```diff
@@ -30,20 +35,86 @@ export function GenomeAssemblyList({ removeSampleFiles = () => {} }) {
...
+   */
+  const updateDefaultGenomeAssembly = (genomeAssembly) => {
+    const { fileInfo: genomeAssemblyObj } = genomeAssembly;
+
+    updateSampleDefaultGenomeAssembly({
+      sampleId: sample.identifier,
+      genomeAssemblyId: genomeAssemblyObj.identifier,
+    })
+      .then(({ data }) => {
+        dispatch(setDefaultGenomeAssembly(genomeAssemblyObj));
```
joshsadam, line 49: Why do you need to store this as state?  The point to using redux toolkit is that it handles the state and you just need to invalidate to update.  It has it's own cache to handle this.

src/main/webapp/resources/js/components/samples/components/GenomeAssemblyList.jsx
```diff
@@ -30,20 +35,86 @@ export function GenomeAssemblyList({ removeSampleFiles = () => {} }) {
...
+        (sample.defaultGenomeAssembly === null && index === 0)
+      ) {
+        actions.push(
+          <Tooltip
+            title={i18n("SampleFilesList.defaultSelectedAssembly")}
+            placement="top"
+            key={`default-tag-tooltip-ga-${genomeAssemblyObj.identifier}`}
+          >
+            <Tag
+              color="#108ee9"
```
joshsadam, line 78: What is this color chosen from?

deepsidhu85: Got this color from the ant design tag component page.It's in one of the examples.

src/main/webapp/resources/js/components/samples/components/GenomeAssemblyList.jsx
deepsidhu85: The default tag doesn't update in the UI without it. If you look in samples.js we invalidateTags: ["SampleDetails"]. However even with this it doesn't update the UI side even when a different assembly is updated. Same thing applies to the setting of a default sequencing object

src/main/webapp/resources/js/components/samples/components/GenomeAssemblyList.jsx
deepsidhu85: I added in css vars for blue based on one of your branches and updated the tag here and in sequenceobjectlist.jsx to use them in aabc9d4a566406d9ef2235054febf0b8a29d4d7f

src/main/webapp/resources/js/components/samples/components/GenomeAssemblyList.jsx
joshsadam: I think it is because the `tag` here is set up wrong.  If you look at the docs for [invalidating tags](https://redux-toolkit.js.org/rtk-query/usage/automated-refetching#selectively-invalidating-lists), we do not actually set the tag type on the object itself. I think where we dropped the ball was, instead of:

```javascript
providesTags: ["SampleDetails"],
```

```javascript
providesTags: (result) =>  [{ type: 'SampleDetails', id: details.identifier }],
```

I have not played with it, but it seems more like what they are looking for and probably why we are not getting invalidation and refetching.

joshsadam (approved):

PR #1337, opened 2022-07-04
Start Galaxy Library timeout when running

### LINKED ISSUE #1338: Timeout when checking status of Galaxy transfers includes queue time
## Describe the bug

This bit of code is used to confirm that files transferred to a Galaxy Library are fully processed by Galaxy before proceeding:

https://github.com/phac-nml/irida/blob/069ae4b76297c1d8dbd286ebb7b65e3ff96d40e8/src/main/java/ca/corefacility/bioinformatics/irida/pipeline/upload/galaxy/GalaxyLibrariesService.java#L210-L241

It works by polling Galaxy for every uploaded dataset and verifies that it's status is `ok`. To avoid infinite waits, I have added a timeout after which the entire analysis pipeline submission in IRIDA will fail.

This status checking code is implemented as a task submitted to a thread pool in Java, which lets me use a call to a Java `Future.get()` to handle the timeout.

https://github.com/phac-nml/irida/blob/069ae4b76297c1d8dbd286ebb7b65e3ff96d40e8/src/main/java/ca/corefacility/bioinformatics/irida/pipeline/upload/galaxy/GalaxyLibrariesService.java#L241

However, calling `Future.get()` with a timeout value includes the time both executing the code to check the status of datasets in Galaxy, as well as the time spent queued in the thread pool in IRIDA (thanks so much to @pvanheus for confirming that the queue time is included here). This becomes an issue if many analysis pipelines are submitted at once, since many may end up in the queue for a long time, leading to a trigger of the `Future.get()` timeout even before the status checking has started running.

1. Submit many analysis pipelines in IRIDA. You may need to increase the limit of the maximum number that are allowed to run. And decrease the `galaxy.library.upload.timeout` to make it easier to trigger this issue.
2. You should eventually get a timeout in uploading files to Galaxy, which switches the IRIDA analysis pipeline to `ERROR`.

## Expected behaviour
I expected the timeout to start when status checking for transferring files to Galaxy starts running.

## Additional context
None.

apetkau:
## Description of changes
Currently, when transferring files to a Galaxy Data Library, the timeout starts when the task is added to a thread pool queue. This means the timeout before file transfer fails in IRIDA includes the time spent in the queue.

This PR changes the Galaxy Data Library transfer code to only start the timeout of data transfers when the status checking code has begun running. This makes use of a [CountDownLatch](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/util/concurrent/CountDownLatch.html).

### Testing

I'm not sure if you wanted to go through testing this, but what I did is upload 59 samples and set the Galaxy upload timeout limit to **15 seconds** and run two scenarios:

* Running all 59 genomes with RefSeq Masher under the old IRIDA - got a bunch of jobs erroring due to upload timeouts
* Running all 59 genomes with RefSeq Masher with IRIDA including these changes - everything works. No errors.

## Related issue
Fixes #1338 

## Checklist


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 ## [22.05.3] - 2022/06/14
 * [Developer]: Fix bug where users with only group access to a project couldn't view sample details or metadata. See [PR 1313](https://github.com/phac-nml/irida/pull/1313)
 * [Developer]: Fix sql order by bug on admin statistics page that was causing page not to load when `ONLY_FULL_GROUP_BY` enabled. See [PR 1313](https://github.com/phac-nml/irida/pull/1313)
+* [Workflow]: Start the Galaxy Data Library timeout when an upload begins rather than when it is first queued up. See [PR 1337](https://github.com/phac-nml/irida/pull/1337)
```
ericenns, line 36: This should be moved to the Unreleased section and not 22.05.3

apetkau: Woops, sorry about that. Fixed in18e1e70d473616a48456153fee07b2a38dcf96bb

ericenns (approved):
👍 Thanks for adding this

PR #1357, opened 2022-08-23
Fixes issue where sample selection is not maintained between table pages.

joshsadam:
## Description of changes
There appears to be a change in the pay redux toolkit returns stat within a slice `addCase` callback, it now appears to return a proxy to the state, which was causing issues comparing objects.   This was causing sample selection across pages to become broken on the project samples table.

I fixed this issue by moving the logic up into an `createAsyncThunk` function where I can get the state directly using the `getState` method.  Instead of just strinifying the object, I updated it to use the lodash isEqual function.

## Related issue
N/A

## Checklist

  - Not required as this is a development issue.

  - Not required an no UI or functional changes

### DIFF
--- a/src/main/webapp/resources/js/pages/projects/redux/samplesSlice.js
+++ b/src/main/webapp/resources/js/pages/projects/redux/samplesSlice.js
@@ -20,6 +20,11 @@ const removeSelectedSample = createAction("samples/table/selected/remove");
 const clearSelectedSamples = createAction("samples/table/selected/clear");
 const clearFilterByFile = createAction("samples/table/clearFilterByFile");
 
+/**
+ * Updates the state of the table filters and search, which triggers
+ * the re-render of the samples table.
+ * @type {AsyncThunk<unknown, void, {}>}
+ */
 const updateTable = createAsyncThunk(
   "samples/table/update",
   async (values, { getState }) => {
--- a/src/main/webapp/resources/js/pages/projects/redux/samplesSlice.js
+++ b/src/main/webapp/resources/js/pages/projects/redux/samplesSlice.js
@@ -3,20 +3,40 @@ import {
   createAsyncThunk,
   createReducer,
 } from "@reduxjs/toolkit";
-import { getProjectIdFromUrl } from "../../../utilities/url-utilities";
+import {
+  getProjectIdFromUrl,
+  setBaseUrl,
+} from "../../../utilities/url-utilities";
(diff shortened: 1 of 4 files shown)


src/main/webapp/resources/js/pages/projects/redux/samplesSlice.js
```diff
@@ -3,21 +3,49 @@ import {
...
+ * the re-render of the samples table.
+ * @type {AsyncThunk<unknown, void, {}>}
+ */
+const updateTable = createAsyncThunk(
+  "samples/table/update",
+  async (values, { getState }) => {
+    const {
+      samples: { options, selected, selectedCount },
+    } = getState();
+    console.log("SEARCH", { state: options.search, values: values.search });
```
deepsidhu85, line 34: Remove the logging statements here and below

src/main/webapp/resources/js/pages/projects/samples/components/SamplesTable.jsx
```diff
@@ -43,33 +43,31 @@ export function SamplesTable() {
...
 
   /**
    * Handle row selection change event
    * @param event
    * @param sample
    */
   const onRowSelectionChange = (event, sample) => {
-    const selected = event.target.checked;
-    if (selected) {
+    console.log(selected);
```
deepsidhu85, line 69: Remove logging statement

deepsidhu85 (changes requested):
This is looking good. Just a couple of things below

src/main/webapp/resources/js/pages/projects/redux/samplesSlice.js
joshsadam: Removed in d73f8bb385198793f63345d399f52a4226e6be79

src/main/webapp/resources/js/pages/projects/samples/components/SamplesTable.jsx
joshsadam: Removed in d73f8bb385198793f63345d399f52a4226e6be79

deepsidhu85: The logging statement is still in this file

deepsidhu85 (changes requested):

joshsadam:
Removed that loggin statement in 643eb5a2eb9d3e8d8f0bd1603af5a8b7ef687f69, thanks

deepsidhu85 (approved):

PR #1371, opened 2022-09-21
Sample details refactor/ metadata fixes

deepsidhu85:
## Description of changes

## Related issue

## Checklist


joshsadam:
Possible for you to update the height of the modal to 80vh?

ericenns (commented):
This functionally works as I would have expected now. The only comment I have is that for really long metadata values should we consider displaying them as a text area instead of a text input. @joshsadam what do you think?

joshsadam:
> This functionally works as I would have expected now. The only comment I have is that for really long metadata values should we consider displaying them as a text area instead of a text input. @joshsadam what do you think?

Definitely agree with that.  Use it since it won't affect short values and all but will make those longer ones so much easier to read.

ericenns (approved):

PR #1386, opened 2022-10-21
Ldap local and sequencer user sign in

JeffreyThiessen:
## Description of changes

Allow local authentication for signing in with non domain users when ldap/adldap is enabled. Used mainly for admin and sequencer users.
Updates to user table to support different user types.

### How to set up LDAP to test
 1. Download (https://directory.apache.org/studio/download/download-linux.html) and setup an ldap server with apache directory studio, this guide explains it step by step to get basic users https://docs.craftercms.org/en/3.0/developers/cook-books/how-tos/setting-up-an-ldap-server-for-dev.html
 2. add a new user to your Directory Studio LDAP server, below we assume it is `jwick`
 3. On Directory Studio, change jwick's password to something you know, like `password`
 4. configure your authentication.configuration file, below is an an example that should work with the above guide
```
# Alternative Authentication modes: ldap, adldap
# Defaults is in memory authentication with bcrypt
irida.administrative.authenitcation.mode=ldap

# LDAP config
irida.administrative.authenitcation.ldap.url=ldap://localhost:10389/
irida.administrative.authentication.ldap.base=dc=example,dc=com
irida.administrative.authenitcation.ldap.userdn=uid=admin,ou=system
irida.administrative.authenitcation.ldap.password=secret
irida.administrative.authentication.ldap.userdn_search_patterns=cn={0},ou=Users
# Optional Argument. change this to 'follow' to handle referrals. See AbstractContextSource javadocs for more info
irida.administrative.authenitcation.ldap.set_referral=ignore
```

### Testing this PR's changes:
While running with ldap enabled (above)
Note: ui changes are handled in a separate PR, so ui will not look correct for domain users
1. If your database has the ldap user `jwick` from testing previous ldap builds, wipe your database or use another username for the follwing, making sure that this new user exists in Directory Studio
2. sign in with a admin user
3. Create a new local user as admin, sign in with new user
4. Create a new sequencer user, use this user to upload data via the irida uploader
6. With Directory Studio running, attempt sign in as `jwick`, a new domain user `jwick` will be created in the irida database.
7. sign out and sign in again as `jwick`, it should authenticate
8. Turn off Directory Studio, attempt sign in as `jwick`, it should not authenticate
9. Verify `user_type` field is being set correctly, 
  a. connect to database `mysql -u test -p` 
  b. `use irida_test;`
  c. `select * from user;`
  d. `jwick` should be `TYPE_DOMAIN`, `admin` and your sequencer user should be `TYPE_LOCAL`

## Related issue
N/A

## Checklist


JeffreyThiessen:
This PR is dependent on https://github.com/phac-nml/irida/pull/1477

src/main/java/ca/corefacility/bioinformatics/irida/config/services/scheduled/ExecutorConfig.java
```diff
@@ -105,7 +105,7 @@ private void addAdminUser() {
 
 		//check if admin user exists.  If not add one
 		if (jdbcTemplate.queryForList("SELECT userName FROM user WHERE userName='admin'").isEmpty()) {
```
ericenns, line 107: Can you add `and user_type='TYPE_LOCAL'` to the query.

src/main/java/ca/corefacility/bioinformatics/irida/ria/web/services/UIUsersService.java
```diff
@@ -152,6 +153,7 @@ public AjaxCreateItemSuccessResponse createUser(UserCreateRequest userCreateRequ
 
 			PasswordReset passwordReset = null;
 			try {
+				user.setUserType(UserType.TYPE_LOCAL);
```
ericenns, line 156: I think this can be removed as the default initializer for `User` already sets this to `TYPE_LOCAL`

src/main/resources/ValidationMessages.properties
```diff
@@ -54,6 +54,7 @@ user.lastName.notnull=You must provide a last name.
 user.lastName.size=Last name must be at least 2 characters long.
 user.phoneNumber.size=Phone number must be at least 4 characters long.
 user.systemRole.notnull=You must provide a system role.
+user.userType.notnull=You must provide a user type.
```
ericenns, line 57: Will this message ever be displayed to the user? I would think not

src/test/java/ca/corefacility/bioinformatics/irida/model/UserTest.java
```diff
@@ -151,6 +151,16 @@ public void testInvalidEmail() {
 		assertEquals(b.getString("user.email.invalid"), constraintViolations.iterator().next().getMessage());
 	}
 
+	@Test
+	public void testNullUserType() {
+		User u = new User();
+		u.setUserType(null);
```
ericenns, line 157: Would it be better to just raise an Exception in `setUserType` if null was passed? Instead of having the null constraint, since this should never happen.

src/main/java/ca/corefacility/bioinformatics/irida/model/user/User.java
```diff
@@ -350,6 +358,10 @@ public void setSystemRole(Role systemRole) {
 		this.systemRole = systemRole;
 	}
 
+	public void setUserType(UserType userType) {
+		this.userType = userType;
```
ericenns, line 362: I made the comment below, but instead of allow null here and catching it with the null constraint, why not just raise an exception here as the code should never allow you to specify null. That way we don't need the translation that should never be shown to the user.

ericenns (changes requested):
Just a few minor comments

ericenns (approved):
Looks good to me!

PR #1403, opened 2022-11-03
HOTFIX: Prevent sharing of locked samples

joshsadam:
## Description of changes
Currently in IRIDA a locked sample can be shared / moved to another project.  This also includes sharing and moving with keeping the locked status.  This PR addresses this by completely removing the ability to share or move locked samples.

## Related issue
N/A

## Checklist


src/main/webapp/resources/js/pages/projects/samples/components/SamplesMenu.jsx
```diff
@@ -144,9 +147,32 @@ export default function SamplesMenu() {
...
     if (selected.size === 0) return;
-    formatAndStoreSamples(`share`);
-    // Redirect user to share page
-    window.location.href = setBaseUrl(`/projects/${projectId}/share`);
+    const { valid, locked } = validateSamplesForMergeOrShare(selected);
+    if (locked.length) {
+      if (valid.length === 0) {
+        message.error(i18n("SampleMenu.share-all-locked"));
+      } else {
+        console.log(valid);
```
deepsidhu85, line 155: Remove logging statement

deepsidhu85 (changes requested):
I still need to test it but the code looks good. Just one change for now.

Also, one thing that I think we may want to address is should a project owner still be allowed to share their samples to their own projects. What if it's the case where a project owner wants to share the samples to another one of their own projects so they can use the sample and write back the results to it, but at the same time want to allow project collaborators to use the sample but not allow them to update any of the sample metadata etc. Since this functionality has been around quite a while we may want to check with some of the users if making this change (preventing the sharing/moving of locked samples) is alright. Thoughts? @joshsadam @ericenns

ericenns:
> that I think we may want to address is should a project owner still be allowed to share their samples to their own projects. What if it's the case where a project owner wants to share the samples to another one of their own projects so they can use the sample and write back the result

The original owner of the samples can go to the original project to share them to other projects. If they deleted the samples from the original project, the samples are now effectively locked and cannot be unlocked, since they would no longer be able to share them (which currently exists). This is why I am ok with this change. I doubt many people were sharing locked samples, I don't think there is anything in the documentation talking about this.

src/main/webapp/resources/js/pages/projects/samples/components/SamplesMenu.jsx
joshsadam: Thanks @deepsidhu85 updated in 9147daf

deepsidhu85:
> > that I think we may want to address is should a project owner still be allowed to share their samples to their own projects. What if it's the case where a project owner wants to share the samples to another one of their own projects so they can use the sample and write back the result
> 
> The original owner of the samples can go to the original project to share them to other projects. If they deleted the samples from the original project, the samples are now effectively locked and cannot be unlocked, since they would no longer be able to share them (which currently exists). This is why I am ok with this change. I doubt many people were sharing locked samples, I don't think there is anything in the documentation talking about this.

Thanks @ericenns. Yeah it that case it definitely makes sense. I will leave it with you. I will test it out shortly but code wise it looks good to me

src/main/webapp/resources/js/pages/projects/samples/components/SamplesMenu.jsx
```diff
@@ -144,9 +147,31 @@ export default function SamplesMenu() {
    */
   const shareSamples = () => {
     if (selected.size === 0) return;
-    formatAndStoreSamples(`share`);
-    // Redirect user to share page
-    window.location.href = setBaseUrl(`/projects/${projectId}/share`);
+    const { valid, locked } = validateSamplesForMergeOrShare(selected);
+    if (locked.length) {
+      if (valid.length === 0) {
+        message.error(i18n("SampleMenu.share-all-locked"));
```
deepsidhu85, line 153: Can we update this to use notifcation instead of message since that's what we use in other places

deepsidhu85 (changes requested):
One other change below

deepsidhu85:
As discussed, it was a little confusing when a user selected both unlocked and locked samples, clicked on the share button, a modal pops up to list the locked samples, then the user clicks ok and it takes you to the share page to continue sharing the unlocked samples. Update it to remove the modal listing the locked samples and rather list these samples on the share -> samples step.

src/main/webapp/resources/js/pages/projects/samples/components/SamplesMenu.jsx
joshsadam: Sure, updated in 80fb51ffff3350dc8955d65e5a8601747345ab7c

src/test/java/ca/corefacility/bioinformatics/irida/ria/integration/projects/ProjectShareSamplesIT.java
```diff
@@ -119,4 +117,35 @@ public void testShareSamplesAsManager() {
...
+
+		LoginPage.loginAsManager(driver());
+
+		// SHARING SINGLE SAMPLE
+		ProjectSamplesPage samplesPage = ProjectSamplesPage.goToPage(driver(), 1);
+		samplesPage.selectSampleByName("sample5fg44");
+		samplesPage.selectSampleByName(LOCKED_SAMPLE_NAME);
+		samplesPage.shareSamples();
+
+		assertFalse(shareSamplesPage.isNextButtonEnabled(), "");
```
deepsidhu85, line 133: Can you add a message here instead of just an empty string

deepsidhu85 (changes requested):

src/test/java/ca/corefacility/bioinformatics/irida/ria/integration/projects/ProjectShareSamplesIT.java
joshsadam: Updated in fabffa8

Also added the missing i18n string in ab8ca1e

deepsidhu85 (approved):

ericenns (approved):
👍

PR #1420, opened 2022-12-02
Removed old reference to `vendor` bundle and external jQuery in webpack

joshsadam:
## Description of changes
Vendor entry was only used to load bootstrap dependencies (required for DataTables) and is no longer required.  Also there was a legacy reference to jQuery in the webpack.config which was removed.  Finally, base.html template was removed as it was no longer required (relative sections moved into page.html)

## Related issue
N/A

## Checklist


src/main/webapp/resources/css/app.css
```diff
@@ -91,42 +91,6 @@ body > .ant-layout {
...
-    }
-
-    & :not(.active) a {
-      color: #717f8c !important;
-      background-color: transparent !important;
-    }
-  }
-}
-
-label.ant-checkbox-wrapper {
```
ericenns, line 126: Was this mistakenly removed, this is set for `.ant-checkbox-wrapper`

src/main/webapp/resources/js/components/Header/MainNavigation/MainNavigation.css
```diff
@@ -39,7 +39,7 @@
 
 .main-navigation .ant-menu-title-content a {
   font-weight: 600;
-  font-size: 1.5rem;
+  font-size: 1.2rem;
```
ericenns, line 42: Why are we reducing the main nav font size here?

ericenns (changes requested):
A few comments, but otherwise looking good.

src/main/webapp/resources/js/components/Header/MainNavigation/MainNavigation.css
joshsadam: When I removed bootstrap, this font in the nav blew up.  This just put it back in check

src/main/webapp/resources/css/app.css
joshsadam: Was there to override bootstraps checkbox css

src/main/webapp/resources/js/components/Header/MainNavigation/MainNavigation.css
ericenns: Ok weird, glad to know that it is the same size pre PR

ericenns (approved):
👍

Issue #1125, opened 2021-11-04
Add method to permanently delete (purge) sequence reads from the filesystem


apetkau:
## Describe your idea for a new feature
Currently in IRIDA there's a method to delete sequence reads (or delete samples from a project). But these don't actually remove the sequence read data from the filesystem. Instead, they hide the sequence read data from appearing in a project/unlink it from a sample.

![image](https://user-images.githubusercontent.com/200517/140355995-8471bbe6-24d0-48fc-8b76-87c33f575880.png)

It would be very useful in some cases to be able to permanently delete sequence reads from the filesystem. Perhaps as a separate operation named something like "Purge" (or "Permanently Delete").

## Describe how you would access this feature?
This could possibly appear as a separate operation available to regular users. Or it could appear as an option available only to administrators (who could maybe have a separate view of all "Deleted/unlinked" data and the ability to permanently delete this data).

## Additional information
None.

apetkau:
An updated screenshot of where deleting sequence files is located now:

![image](https://user-images.githubusercontent.com/200517/216126696-6eaf35a0-a4ad-442e-bf5d-7e823f34d3c9.png)

ksierks:
![image](https://user-images.githubusercontent.com/325703/216365156-04d07fa2-5696-4cde-88da-200e6d7f918a.png)

From doing a quick test, I believe if you remove a sample from a project or remove a sequencing run, it deletes those records from the database. In both cases, they leave behind their sequencing object and files. We need a way to clean up sequencing objects.

ksierks:
- deleting a sequence run deletes all the database records (sequencing_run, sequencing_object, sequence_file, samples, etc), but not the actual file on the file system
- deleting a sample with sequence data, deletes the sample from the database, but not the sequencing_object

PR #1469, opened 2023-02-16
Delete sequence files from file system via cleanup script

ksierks:
## Description of changes
Created a script to delete sequence files from local file system that have already been previously deleted from the database.
Previously deleted sequence files are retrieved from the audit table, because of the inconsistent way they are removed within IRIDA.
The intention is for this script to be run once to do an initial cleanup the file system. There will be subsequent PRs addressing ongoing file clean up. 

#### Steps to test
- create a new sequence run via the irida_uploader
- login to IRIDA as admin and delete the newly created sequence run
- run the purge_sequence_files script.py

## Related issue
https://github.com/phac-nml/irida/issues/1125

## Checklist


src/main/resources/scripts/sequence-files/purge_sequence_files.py
```diff
@@ -0,0 +1,48 @@
...
+def list_sequence_files(host, user, password, database):
+    db = mysql.connector.connect(
+        host=host,
+        user=user,
+        password=password,
+        database=database
+    )
+    cursor = db.cursor()
+    # TODO: Should we double check this file doesn't exist in the actual table in case it was manually restored?
+    cursor.execute("SELECT DISTINCT file_path FROM sequence_file_AUD WHERE revtype=2")
```
JeffreyThiessen, line 15: I see a big issue here. The revision number does not dictate what type of file it is.
With this line I assume you are trying to delete just the unzipped files, as they typically exist as type 2, and the .gz files exist as type 1.
This is not always the case, in this image you can see I did the click and drag to upload 2 files, file_1.fastq, and file_2.fastq.gz
image

JeffreyThiessen (changes requested):

src/main/resources/scripts/sequence-files/purge_sequence_files.py
JeffreyThiessen: In the log here, you can see that file_1.fastq is under /55/1/ , file_2.fastq.gz is in /56/1/ and the unzipped file_2.fastq is in /56/2/

ksierks: I think there might be some confusion between REVTYPE and file_revision_number. My query is getting all the deleted files from the audit table.
You're right, I missed the compressed files. I think I had tunnel vision. I've updated the query to include all the file paths (compressed & uncompressed) that have been deleted. 
I also added an argument to the script to specify the sequence file base directory. 
Please see 47c5f00214b570cd20f2a15992b1472ebb104185.

deepsidhu85:
@ksierks Hi Katherine, I see that the script cleans up files on the local file storage system but not from cloud storage (aws, azure). Are you planning on adding that in here or in another PR?

ksierks:
> @ksierks Hi Katherine, I see that the script cleans up files on the local file storage system but not from cloud storage (aws, azure). Are you planning on adding that in here or in another PR?

Hi Deep. I wasn't planning on it, unless told otherwise. The file system has 10+ years worth of files. IRIDA hasn't been installed in the cloud that I know of. The intention is to only run the script once and have IRIDA cleanup files.

deepsidhu85:
> > @ksierks Hi Katherine, I see that the script cleans up files on the local file storage system but not from cloud storage (aws, azure). Are you planning on adding that in here or in another PR?
> 
> Hi Deep. I wasn't planning on it, unless told otherwise. The file system has 10+ years worth of files. IRIDA hasn't been installed in the cloud that I know of. The intention is to only run the script once and have IRIDA cleanup files.

Sorry about that Katherine. I missed the part in the description where the ongoing cleanup will be addressed in subsequent prs. Disregard my above comment! :)

deepsidhu85 (approved):
Looks good to me!

src/main/resources/scripts/sequence-files/purge_sequence_files.py
```diff
@@ -0,0 +1,61 @@
...
+    result = cursor.fetchall()
+    cursor.close()
+    db.close()
+    return result
+
+def main():
+    parser = argparse.ArgumentParser(description="This program lists the sequence files and folders that have been previously deleted in IRIDA.")
+    parser.add_argument('--purge', help="Deletes the sequence files and folders from the filesystem.", action="store_true")
+    parser.add_argument('--baseDirectory', help="The sequence file base directory.", required=True)
```
JeffreyThiessen, line 38: Please add a line to the help saying something like "the folder is usually formatted like `.../irida/sequence-files/`
I wasn't sure if I just needed to point it to /tmp/irida/ or where when I used it the first time.

src/main/resources/scripts/sequence-files/purge_sequence_files.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/python
+import argparse
+import mysql.connector
+import os
+
+def remove(path, purge):
+    if purge:
+        try:
+            if os.path.isdir(path):
```
JeffreyThiessen, line 9: Can you add a check that the path exists first? When it tries to delete a directory/path that is already gone (for example, removed via admin panel in other branch) it only says `Unable to delete`. This might confuse someone why some files are not deleting, when they didn't exist in the first place.
image

JeffreyThiessen (changes requested):
Looks good. Just a few small changes.

src/main/resources/scripts/sequence-files/purge_sequence_files.py
ksierks: Please see 2dd698774e39cd05609d0e3f69ee472c31be573b.

src/main/resources/scripts/sequence-files/purge_sequence_files.py
ksierks: I added a default value for the sequence directory. I hope that's okay. Please see 1b0070c416a6388551174d90069a43ee946862eb.

src/main/resources/scripts/sequence-files/purge_sequence_files.py
```diff
@@ -0,0 +1,61 @@
...
+
+def remove(path, purge):
+    if purge:
+        try:
+            if os.path.exists(path):
+                if os.path.isdir(path):
+                    os.rmdir(path)
+                elif os.path.isfile(path):
+                    os.remove(path)
+                else:
```
JeffreyThiessen, line 14: This else might have the wrong indentation? I don't think it can be reached and it was intended to be fore the if exists line.

JeffreyThiessen: I would probably be a good idea to add a print statement when they get removed too, because it just says "All done." when it removes files, so if you mistype your directory and run it, we have no idea if the files were deleted or not before manually checking (this happened to me while testing it)

ksierks: Please see e77dc42b0bcf2e25ac53a3ebfb4e9ffa905ecf20.

JeffreyThiessen (approved):
Looks good

PR #1352, opened 2022-08-09
Displaying identifiers for project drop downs

ksierks:
## Description of changes
What did you change in this pull request?  Provide a description of files changed, user interactions changed, etc.  Include how to test your changes.

## Related issue
#1347

## Checklist
Things for the developer to confirm they've done before the PR should be accepted:

* [x] CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change.
* ~[ ] Tests added (or description of how to test) for any new features.~
* ~[ ] User documentation updated for UI or technical changes.~

PR #1474, opened 2023-03-03
Add project id to project synchronization dropdown

ksierks:
## Description of changes
Added project id to project dropdown.

image

## Related issue
An extension of #1352

## Checklist

### DIFF
--- a/src/main/webapp/resources/js/types/irida/index.d.ts
+++ b/src/main/webapp/resources/js/types/irida/index.d.ts
@@ -131,6 +131,7 @@ declare namespace IRIDA {
     analysisPriority: string;
     users: User[];
     analysisTemplates: string[]; // TODO (Josh - 6/7/22): What should this be
+    url: string;
   }
 
   type ProjectMinimal = Pick<Project, "id" | "name">;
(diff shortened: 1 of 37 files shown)


src/main/webapp/resources/js/types/irida/index.d.ts
```diff
@@ -131,6 +131,7 @@ declare namespace IRIDA {
     analysisPriority: string;
     users: User[];
     analysisTemplates: string[]; // TODO (Josh - 6/7/22): What should this be
+    url: string;
```
joshsadam, line 134: Do we want to name this `remoteUrl` to directly map to the server Project model?  That is what it is used for right?

joshsadam (changes requested):
Just one comment, and it is a question.  If it is the remoteUrl I think we should name it that.

src/main/webapp/resources/js/types/irida/index.d.ts
ksierks: Ok, I was just doing what was easiest for me in the moment. Please see b9fae817b6f651d44e6e9ada08fd052af4814eef.

joshsadam (approved):
This looks good, thanks

PR #1481, opened 2023-03-29
Ldap UI changes

JeffreyThiessen:
## Description of changes
* Blocking UI elements for modifying user objects that are controlled by domain (ldap/adldap)
* first name, last name, email, phone number and enable/disable checkbox are all greyed out on UserDetailsPage (for both admins and current user)
* change password form is disabled on UserChangePasswordForm for both admins and current user.
* Added text on UserDetailsPage and UserChangePasswordForm indicating that disabled fields are managed by ldap/adldap
* Disabled enable/disable checkbox on UserTable for domain users.
* Added UserType column to UserTable to indicate between local and domain users

## Related issue
N/A

## Checklist


src/main/webapp/resources/js/pages/user/components/UserChangePasswordForm.jsx
```diff
@@ -42,16 +46,20 @@ export function UserChangePasswordForm({ userId, requireOldPassword }) {
   };
 
   return (
-    <>
+    <Skeleton loading={isLoading}>
       <Typography.Title level={5}>
         {i18n("UserChangePasswordForm.title")}
       </Typography.Title>
+      <Typography.Text type="secondary" hidden={!userDetails.domainAccount}>
```
joshsadam, line 53: **chore (blocking):** Instead of just hiding this, can you wrap it in an if statement so that it never gets rendered to the page.

joshsadam: Also on the `UserSecurityPage` add a condition to the if statement on line 25 since we do not need to see the password policy if the user is LDAP

joshsadam:
On the Admn > Users > account page, make sure you hide the password policy and ability to reset a users password since that should all be handled by LDAP

joshsadam (changes requested):
Overall, I really like the way this is looking and functioning.  Just a couple changes I would like to see.  Mostly just hiding the password policy and reset button.

src/main/webapp/resources/js/pages/user/components/UserChangePasswordForm.jsx
JeffreyThiessen: Changes added in https://github.com/phac-nml/irida/pull/1481/commits/93f0836065f94f1f59d46f14dcdaad55fc8b2258

joshsadam: On the admin > list > account page, thank you for removing the alert with the password requirements, I can still see the reset password stuff though, can you please also remove that.

JeffreyThiessen: Oh I missed that because I didn't have email configured. Should be fixed now. [f49faa5](https://github.com/phac-nml/irida/pull/1481/commits/f49faa549e1edb21051a40c3c1cb6c2bfc071d1a)

joshsadam (approved):
Got it working :), Thanks Jeff, this looks great

PR #1484, opened 2023-04-14
hotfix/sync old irida

apetkau:
## Description of changes

Fixes issue with syncing with old IRIDA (specifically with 20.01). Specifically, this explicitly changes the request headers for the API to respond with JSON. Previously, in some instances, we were defaulting to requesting XML, which is not supported by the older IRIDA instances. This would have first occurred in https://github.com/phac-nml/irida/commit/b334e59169f926e8fa9e41754806c0dbdc8f4a55 where we disabled XML support in the REST API.

This should be merged into `main` after the branch `fix-remove-template-field`.

This can be tested by setting up two instances of IRIDA (one 20.01 and another this version) and attempting to sync from newer IRIDA to 20.01. Without this fix, it will fail. With this fix it will succeed.

## Related issue

## Checklist


ericenns (approved):
Looks good to me!

PR #1504, opened 2023-12-07
Fixed sort on project members table

joshsadam:
## Description of changes

Fixed issue sorting by name on the project members table.  

When the method was called in the service it the sortColumn was set as `name` as that's what we have in the front end (combines the first and last name) which was retrieved on getting the members for the page originally. When we either sorted on the name column or switch pages it called the method with the sort params for the column as we set the sorter for it on the column in the front end code. When it went to do the order by on the query it was trying to get the name from the projectuserjoin which doesn't have a name column, neither does the User model, hence why if the column is name we sort on the first name for the user.

## Related issue

Link to the GitHub issue this pull request addresses using the `#issuenum` format. If it completes an issue, use `Fixes #issuenum` to automatically close the issue.

## Checklist

*   [ ] CHANGELOG.md (and UPGRADING.md if necessary) updated with information for new change.
*   [ ] Tests added (or description of how to test) for any new features.
*   [ ] User documentation updated for UI or technical changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
-# [23.10.1] = 2023/12/07
+# [23.10.1] - 2023/12/07
 * [UI]: Fixed issue where filter inputs required focus when they are opened on the project samples page. See [PR 1503](https://github.com/phac-nml/irida/pull/1503)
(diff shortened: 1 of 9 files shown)


deepsidhu85 (changes requested):
Looks good just one comment below

deepsidhu85 (approved):

PR #1580, opened 2026-03-14
docs: add Tomcat9 sandbox writable dirs config for Ubuntu package

### LINKED ISSUE #1555: Add documentation on writable directories for tomcat9 in Ubuntu
Tomcat9 in Ubuntu is configured by default to restrict which directories are writable (it is sandboxed). This will cause issues when checking for write permissions on the `sequence-files`, `output-files,` etc directories. This can be changed via configuration settings in tomcat.

See for example https://stackoverflow.com/a/56835997

This feature request is to add a note in the install documentation on how to configure different directories as writable for Tomcat9 in Ubuntu.

## Additional information
None.

Abdeltoto:
Closes #1555

When Tomcat9 is installed via apt on Ubuntu/Debian, it is sandboxed by systemd and only has write access to a few default directories. IRIDA needs write access to sequence-files, output-files, reference-files, and assembly-files.

This PR adds documentation explaining how to configure \ReadWritePaths\ in a systemd override so IRIDA can write to these directories when using the Ubuntu Tomcat9 package.

Made with [Cursor](https://cursor.com)


apetkau:
Thank you for your contribution, but we are no longer doing development work on this repo.

